PR #162, opened 2014-12-02
Fixes #159: PHPUnit tests tuning

eps90:
Hey,
As I mentioned in #159 , I've done some fixes and corrections for PHPUnit test environment for Magallanes. I've also noticed and fixed following problems:
* lack of `phpunit` in `bin` directory while doing installing dependencies by composer. I've fixed that by adding `bin-dir` to `composer.json`
* no colors... for phpunit output :) Changed phpunit default configuration in `phpunit.xml.dist`
* coverage - I've added `coversDefaultClass` and `covers` in test class to be sure that those test methods test that what we want to test

Please free to code-review!
@SenseException @andres-montanez 

PS. Sorry, I had to close previous PR, because I accidentally compared it to `master` branch again :(

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 vendor
 mage.phar
+bin
+!bin/mage
 
 # OS generated files # // GitHub Recommendation
 ######################
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,9 @@
             "Command\\": ".mage/commands"
         }
     },
+    "config": {
+        "bin-dir": "bin"
+    },
     "bin": [
         "bin/mage"
     ]
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "a19528b890d301384e45c1ed7d221e26",
+    "hash": "c66ae8cd7e44d614445b273f310d9c34",
     "packages": [],
     "packages-dev": [
         {
@@ -755,6 +755,7 @@
     "minimum-stability": "stable",
     "stability-flags": [],
     "prefer-stable": false,
+    "prefer-lowest": false,
     "platform": {
         "php": ">=5.3"
     },
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -5,8 +5,8 @@
          backupGlobals="false"
          verbose="true"
          strict="true"
-         colors="false">
-    <testsuite name="small">
+         colors="true">
+    <testsuite name="unit-tests">
       <directory suffix="Test.php">tests</directory>
     </testsuite>
 
--- a/tests/MageTest/Console/ColorsTest.php
+++ b/tests/MageTest/Console/ColorsTest.php
@@ -11,6 +11,7 @@ use PHPUnit_Framework_TestCase;
  */
 class ColorsTest extends PHPUnit_Framework_TestCase
 {
+    protected $noColorParameter = "no-color";
     /**
      * @group 159
      * @covers ::color
@@ -20,7 +21,7 @@ class ColorsTest extends PHPUnit_Framework_TestCase
         $config = $this->getMock('Mage\Config');
         $config->expects($this->once())
             ->method('getParameter')
-            ->with('no-color')
+            ->with($this->noColorParameter)
             ->will($this->returnValue(false));
 
         $string = '<green>FooBar</green>';
@@ -41,7 +42,7 @@ class ColorsTest extends PHPUnit_Framework_TestCase
         $config = $this->getMock('Mage\Config');
         $config->expects($this->once())
             ->method('getParameter')
-            ->with('no-color')
+            ->with($this->noColorParameter)
             ->will($this->returnValue(true));
 
         $string = '<black>FooBar</black>';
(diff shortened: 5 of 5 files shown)


eps90:
Up

eps90:
Added one more commit, disabling strict testing. In this stage, we don't need to be reported of risky test. It will be helpful in bigger refactor though.

eps90:
@SenseException I've noticed you use `@group` annotations. Are they needed? Do they have some special meaning in testing?

SenseException:
Group can help you to only execute specific tests, that belong somehow together. Many projects use the issue id as group name when new features are added or a bug was fixed, so only the belonging tests can be executed. Mostly they are used for methods but it doesn't hurt to add groups to a class too.

We should keep the strict mode, since it would be troublesome to change it back later to strict mode and existing tests start to fail. Strict also make tests "fail", that haven't any assertions and do actually nothing. Tests also fail, if they are taking to much time, which can be labeled with other annotations in case the runtime is expected. Unexpected output should also make a test fail.

group-annotation: https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.group
strict mode: https://phpunit.de/manual/current/en/strict-mode.html

eps90:
Ok, I've removed that commit. But when I tried to run sample unit tests yesterday, I had a long long list of warnings that will be shown as long as we won't refactor this code.

eps90:
I've just pushed fix commit due to someone else edited the `composer.lock` and there were conflicts.
@SenseException Can you take a look and tell is the whole fix is ok?

SenseException:
Right now, we have only one real requirement in Composer, excluding the PHP version. PHPUnit has a fixed version number in composer.json but maybe this is going to change. Right now, prefer-lowest has no big impact. This can be discussed in case Magallanes has to work with a minimum version for one or more non-dev requirements that may be added in the future.

But maybe we should use prefer-stable for composer from the beginning.

What kind of conflicts happend?

eps90:
@SenseException I meant you can take a look at whole fix, not only last commit :)
Well, `prefer-lowest` was not added by me. It has done automatically while doing `composer update`. The problem was that someone has mastered with `composer.lock` and his PR has been merged (and nothing changed in `composer.json` thought. Someone has installed dependencies by `composer update`?!). The new `composer.lock` landed in `develop` branch. I had a different, updated `composer.lock` with information about `bin-dir`, where to put dependencies' binaries and executables. And there it is, the conflict's ready!
Please take a look at tests class is that form acceptable. If you approve it, leave a message, like upvote - :+1: (it's the way we do in my fulltime job ;) )

SenseException:
Those Improvements look good to me. Cover- and Group-Annotation in tests should be recommended for future contributions.

There was the short discussion about `no-color`, actually config keys, being in a constant.
Properties should be in private scope since we don't want to encourage inheritance like they do in Zend Framework. It is a good practice to start with the lowest visibility.
Single quotes are preferable for strings in PHP and maybe we should add that to a guideline too (#174).

eps90:
@SenseException Private variables would be a good idea here but it's not necessary though. Of course I can change it but I think it does not affect the security or encapsulation, since we don't extend any tests now. So yes, it's quite redudant here but in the other hand it's kind of habit. 
Quoting strings style convention is quite controversial topic. Especially when you use string interpolation. There is no big difference in performance (there was.. in PHP 4) So here we can give developers a free hand. In contribution quidelines we should restrict readbility - quoting doesn't affect the readability that much :)

SenseException:
I'm trying to insert some consistency here. Using variables in bigger strings may be harder to read.

I prefer

```php
echo 'Hello ' . $bar . ', how are ' . $foo->foo() . PHP_EOL;
```
more than

```php
echo "Hello $bar, how are {$foo->foo()}" . PHP_EOL;
```

How good highlighting is depends on the IDE. Functions like sprintf are even better too read.

You're writing "we", which is the correct term for a project, where everyone is allowed to contribute. But "we" is not a constant. Contributors will look into current code and copy its patterns from the other users. If there is no real pattern visible because everyone is coding different, the code is going to be more messy.

Nobody will know every conversation and ticket just like I missed the contribution guidelines ticket #183. But the `protected`scope is there and it allowes to use that property with inheritance. FCol becomes more popular while inheritance is going to be more and more some kind of anti-pattern. I kind of like the extreme idea of @ocramius his blog post http://ocramius.github.io/blog/when-to-declare-classes-final/ even though it is kind of ~~Sparta~~ madness for everyone not familiar with it.

eps90:
You wrote "I prefer", so it's actually the way you like to write in PHP. But for your example, clean code is not like:
```php
echo "Hello $bar, how are {$foo->foo()}" . PHP_EOL;
```
but:
```php
$fooValue = $foo->foo();
echo "Hello $bar, how are $fooValue";
```
In the other hand, forcing single quotation you forbid also using special characters, like `\n`, `\t` and so on, in your strings. So one dev will prefere single quotation, while another, double. Personally I prefer double quoting because in other, C-like languages double quoting means string, while single - a character. Since I code apps not only with PHP, it's my own habit to do that.

Of course, I agree with using `protected` keyword here. But I think the main reason is to set this property only for this test, no matter if it's gonna be a base class for another test. I'll change it ASAP.

eps90:
@SenseException But of course if you have any notes for contribution process, as well as for code style, all feedback is welcome :)

SenseException:
Contribution process thoughts will be added to #183 or related tickets. My overview of tickets is not as good as yours. :-)

Without double quotes, `{$foo->foo()}` even cannot be used. Just by telling that single quotes should be used, we don't need to reference to a few clean code topics. PHP_EOL is the constant for a new line, even though it is not necessarily the same as `\n`. I don't know if there's a constant for tab.

eps90:
@SenseException Great, add your ideas there :) I've added another commit that changes that variable visibility. Could you take a final look to whole PR and tell if it's ok now ? If yes, I'm gonna merge it :) And testing process can go on :)

SenseException:
@eps90 I say there is no real blocker left and it can be merged. :sunny:

eps90:
@SenseException Great! thanks for great feedback! :)

PR #166, opened 2014-12-08
Feature/create directories

komex:
Make target dir before invoking "ln".
Fix bug with spaces in path.


andres-montanez:
Hi! Thanks for your interest in the project and the PR!

PR #171, opened 2014-12-13
Fix/Scm working directory

jhuet:
git local commands should be executed in the deployment:from directory.


andres-montanez:
Thanks!

PR #172, opened 2014-12-23
fix: DeployCommand does not fail after failed "Deployment Task"

dmotylev:
fix: DeployCommand does not fail when no one "Post-Deployment Tasks" failed and failed "Deployment Task" exist


andres-montanez:
Thanks!

PR #173, opened 2014-12-28
Fix/composer autoload

jhuet:
@eps90 i'd like your opinion on this because it seems that you added composer autoloading in the ``bin/mage`` and ``composer.json`` files.

On my environment composer autoloader was never called because the ``__DIR__ . '/../vendor/autoload.php'`` path was leading on no file because ``__DIR__`` is referencing this directory : ``vendor/andres-montanez/magallanes/bin``. So i changed it according to that.

Then, because i wasn't using the internal Mage autoloading anymore but Composer's i encountered the same problem for userland ``Task`` and ``Command`` namespaces because the .mage directory is not correctly referenced. Thus i had to change the PSR-4 autoload directories in ``composer.json``

I guess this has never been spotted because everybody's using Mage with another application that is requiring Composer's autoloading file on its side ?

### DIFF
--- a/bin/mage
+++ b/bin/mage
@@ -16,7 +16,9 @@ $baseDir = dirname(dirname(__FILE__));
 define('MAGALLANES_VERSION', '1.0.3');
 define('MAGALLANES_DIRECTORY', $baseDir);
 
-if (file_exists(__DIR__ . '/../../../autoload.php')) {
+if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
+    require_once __DIR__ . '/../vendor/autoload.php';
+} else if (file_exists(__DIR__ . '/../../../autoload.php')) {
     require_once __DIR__ . '/../../../autoload.php';
 } else {
     require_once $baseDir . '/Mage/Autoload.php';
--- a/bin/mage
+++ b/bin/mage
@@ -16,8 +16,8 @@ $baseDir = dirname(dirname(__FILE__));
 define('MAGALLANES_VERSION', '1.0.3');
 define('MAGALLANES_DIRECTORY', $baseDir);
 
-if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
-    require_once __DIR__ . '/../vendor/autoload.php';
+if (file_exists(__DIR__ . '/../../../autoload.php')) {
+    require_once __DIR__ . '/../../../autoload.php';
 } else {
     require_once $baseDir . '/Mage/Autoload.php';
     $loader = new \Mage\Autoload();
--- a/composer.json
+++ b/composer.json
@@ -11,8 +11,8 @@
     "autoload": {
         "psr-4": {
             "Mage\\": "./Mage",
-            "Task\\": ".mage/tasks",
-            "Command\\": ".mage/commands"
+            "Task\\": "../../../.mage/tasks",
+            "Command\\": "../../../.mage/commands"
         }
     },
     "bin": [
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "andres-montanez/magallanes",
+    "name": "jhuet/magallanes",
     "description": "A Deployment Tool for PHP Applications",
     "homepage": "http://magephp.com",
     "license": "MIT",
--- a/composer.json
+++ b/composer.json
@@ -11,8 +11,8 @@
     "autoload": {
         "psr-4": {
             "Mage\\": "./Mage",
-            "Task\\": "../../../.mage/tasks",
-            "Command\\": "../../../.mage/commands"
+            "Task\\": [".mage/tasks", "../../../.mage/tasks"],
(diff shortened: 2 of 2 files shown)


eps90:
Hi @jhuet ,
Sorry for the delay but during vacation I was not working almost at all.
You're right and I feel so awkward and ashamed now...
Stupid mistake that I didn't noticed because, as you said, I was testing the autoloader on pure Magallanes repository, not as a project dependency. But let me point few things:

1. You've made a PR to `master` branch, while all PR should go into `develop` branch (see `README` in repo's root)
2. Requiring or including `autoloader.php` should check whether `__DIR__ . '/../vendor/autoload.php'` or `__DIR__ . '/../../../autoload.php'` exists, I think.

I think there's also another problem - your PSR configuration makes Magallanes impossible to run tasks or commands living in magallanes' project root. I've made a quick research and I've found that PSR4 allows to set multiple paths for given namespace prefix: https://getcomposer.org/doc/04-schema.md#autoload

Tell me what do you think. If you want me to pay for my faults, I can make a PR into your fork with fixes.

jhuet:
Hi @eps90, no problem for the delay, vacations are good because they indeed don't involve working at all ;)

1. Oops, my turn to not pay enough attention...
2. Here i'm not sure i'm getting what you're saying. I don't see in what context `__DIR__ . '/../vendor/autoload.php'` could be working ?
3. There too, i feel like there's another way of using Magallanes that i'm not familiar with. What directories and what namespaces are you refering to when talking about magallane's project root ?

No problem for me to fix this don't worry ;)

eps90:
Sorry for inconsitency in my answer :) By project root I mean the root of Magallanes itself. 
Ad 2a: I mean the case when you start magallanes straight from project root. For me it's useful working with the code. Then I don't have to install Magallanes to target project to test if it's working.
Ad 2b: In the future it can be useful also for unit tests runner.
Ad 3. I think it's quite obvious now. The sample composer's PSR4 configuration would make able to run tasks from Magallanes itself as well as from target project, from the same namespace. 

Am I thinking right or is it too much complication for such type of project? What do you think?

jhuet:
So you're saying that in your scenario, you're having `__DIR__` referencing the [`Mage`](https://github.com/andres-montanez/Magallanes/tree/master/Mage) directory ?

I think i'm trying to be as lazy as you in my scenario by not having Magallanes in every single project i want to deploy to. What i do though is to create an empty directory with just a `composer.json` file in it requiring Mage and install it. Thus i just have Magallanes in `vendor/andres-montanez/magallanes` and i create a `.mage` directory for my configurations, tasks, commands...

Now if i wanna code / commit on Mage, i just checkout it in another directory and replace the `vendor/andres-montanez/magallanes` with a symlink to where i checkout Mage. Thus i have the exact same environment as anybody using it with composer.

Does that make sense or am i totally out of subject ? :)

eps90:
Something like that :) but my workflow, like many others, is a bit different.
I like to commit changes as often as possible to have project history and in case of mistakes to revert only single commits, that is small portions of code (I hope I told that clearly enough :) )

But in the case when there will be no composer autoloader for the Magallanes itself and you'll be trying to run unit tests, you run them with different, core autoloader.
* First thing, it's a smell. 
* Second - results can be different in target application (composer autoloader) and in the project itself (core autoloader).
* Third - in case of running tests (unit or even manual) you can't test actual state of application because it behaves a bit differently than in production.

You get it now what I think about? :)

jhuet:
Alright i guess i get it now. So basically, what's left to do is to be able to load composer's autoloader in `bin/mage` in both situations depending on wich file exists : 

``` php
  if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
      require_once __DIR__ . '/../vendor/autoload.php';
  } else if (file_exists(__DIR__ . '/../../../autoload.php')) {
      require_once __DIR__ . '/../../../autoload.php';
  } else {

```

And we have to allow the use of `Task` and `Command` namespaces in both directories by changing the PSR-4 autoload section of `composer.json` thanks to [what you found](https://getcomposer.org/doc/04-schema.md#autoload) in composer's doc :

``` json
  "Task\\": [".mage/tasks", "../../../.mage/tasks"],
  "Command\\": [".mage/tasks", "../../../.mage/commands"]
```

If you agree, i'll just do that _blindly_ without switching back to my vendor for testing if you don't mind ;)

eps90:
This is excatly what I was talking about :)
Sure, write that, I can test that by forking your repo because I have already ready an environment to do that 
:)

andres-montanez:
Hi everyone!
Thanks a lot for this development.
So, is the PR ready to be merged?
I don't mind, in this case, doing the merge on master.
Let me know!

Regards.

jhuet:
I'd like @eps90 to review / test it before we merge this to see if both ways are working.

eps90:
@andres-montanez @jhuet I've already checked it, everything is fine :)

andres-montanez:
Many thanks!

PR #175, opened 2015-01-02
Test added for VersionCommand

claudioflagbit:
I've added a test for the VersionCommand class, which is a pretty simple command. But because of the static methods, that are used and executed by invoking `VersionCommand::run()` this test got a bit ugly by using the Reflection-API to disable logging (uses filesystem) and setting `Config`.

@andres-montanez: It would be a BC break if we refactor the static methods so that we are able to mock dependencies. But a simple class like VersionCommand becomes more of an unstable test, if it executes parts of code, that should be mocked and injected.

Would you consider something like a Milestone-Version of Magallanes, that contains refactoring?


eps90:
Whoa, so much of workarounds only to test if console outputs correct string. I think until we use statics we won't overcome problems such as these. 
But if I can suggest something... injecting constants by PHPUnit mechanism, maybe there should be some helper method in `VersionCommand` class to return this contant's value? Then you can easily mock that method. I'm saying this because I can see two smelly things:

1. Unit tests should be indepentent from everything outside it, so you have set the contant in configuration but you actually don't test the class itself but the constant that is set in entry point of application.
2. You shouln't use that constant you are dependent on in assertion. So instead of 
```php
$this->expectOutputString('Running Magallanes version ' . MAGALLANES_VERSION . str_repeat(PHP_EOL, 2));
``` 
you should write
```php
$this->expectOutputString('Running Magallanes version 2' . str_repeat(PHP_EOL, 2));
```
Please tell me what do you think.

eps90:
And one more thing - if we're keeping the convention of creating test classes with familiar structure as production code, you've made a typo in class path (and namespace, consequently) - there should be `BuiltIt`, not `BuildIn`.

claudioflagbit:
BuildIn is the namespace used by the Magellanes-Commands and for better readability and consistency of the tests, the same namespace should be used by its test-class.

There shouldn't be any problem with the usage of the constant, but I guess you're right about more independent tests.
I've also considered to move these constants to a class instead of being created in the mage-cli, but made another branch for that. In this mentioned branch, the version inside the tests would change with every version and using the constant inside tests will become important. I did no PR for that at the moment.

eps90:
> BuildIn is the namespace used by the Magellanes-Commands and for better readability and consistency of the tests, the same namespace should be used by its test-class.

Exactly! :) Take a look in your test class, the name after `namespace` keyword and the other name, after `use` keyword with production class.
Your test class' full name is:

    MageTest\Command\BuildIn\versionCommandTest

While original class is:

    Mage\Command\BuiltIn\VersionCommand

claudioflagbit:
Now I see. I was searching for `BuiltIt` like you wrote and didn't find anything. :-)
I'll fix this.

tests/MageTest/Command/BuiltIn/VersionCommandTest.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $refClass = new \ReflectionClass('Mage\Console');
+
+        $refProperty = $refClass->getProperty('logEnabled');
+        $refProperty->setAccessible(true);
+        $refProperty->setValue(false);
+
+        $config = $this->getMock('Mage\Config');
+        $config->expects($this->once())
+            ->method('getParameter')
```
eps90: `->with(...)`
Expectations should be more restrict, especially for `Config::get` because it's been using in so many places that your tests can be inconsistent.

SenseException:
After #197 is merged, this PR can use the BaseTest test helper class.

eps90:
@SenseException As you mentioned you can change your test to use `BaseTest` class :) Correct, please, the thing I mentioned some time ago and this can be merged too.

SenseException:
Changed to BaseTest.

eps90:
Great, thanks! ;)

PR #176, opened 2015-01-05
Task factory test

claudioflagbit:
Like the Command Factory tests in #167, I created unit tests for the Task Factory. I also did minor changes in the Factory itself.


tests/MageTest/Task/FactoryTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->config = $this->getMock('Mage\Config');
+    }
+
+    /**
+     * @dataProvider taskDataProvider
+     */
+    public function testGet($taskData)
+    {
+        $task = Factory::get($taskData, $this->config);
+        $this->assertInstanceOf('\\Mage\\Task\\AbstractTask', $task);
```
eps90: Shouldn't you check also if given task is instance of the Task's class?

tests/MageTest/Task/FactoryTest.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * @dataProvider taskDataProvider
+     */
+    public function testGetWithOptionalParams($taskData)
+    {
+        $task = Factory::get($taskData, $this->config, true, 'production');
+        $this->assertInstanceOf('\\Mage\\Task\\AbstractTask', $task);
```
eps90: As above

eps90:
1. I've added few comments for your code
2. Please, make sure your file ends with one empty line (PSR2 restriction)
3. Rebase your branch with `develop`, it looks like you have some conflicts
4. I'm not sure and I'm thinking about it for some time - is it necessary to create those "dummy" classes for testing purpose? Unit tests should go on some low-level layer. To consider :)
5. It's strange but your tests covers also `AbstractTask`'s constructor :)

SenseException:
There may be some connection to #167 because of usage of substitute classes.

eps90:
@claudioflagbit Please look at #167 PR for some directions for avoiding those "dummy" classes. And once again, rebase/merge with `develop` - I'm 99% sure you have some conficts with `composer.lock`. Add also PHPUnit's `@coverage` annotations to prevent covering other classes than `Task\Factory`.

SenseException:
When #167 is merged, its changes can also be adopted for the task factory to keep both tests consistent since they have the same mechanics.

eps90:
@claudioflagbit ping ;) Are you gonna do something with this PR?

SenseException:
@eps90 Yes, I'm going to adapt the tests.

eps90:
Haha no way, it's your second account ? :D Great then :) I've got a package with some commands tests locally (tested about 5 command). Maybe I'll create a PR with them (should I?).

SenseException:
Task tests are now without dummies.

eps90:
All right. Tests are passing, coverage is 100%... almost.
This is what I get when I ran PHPUnit on your branch:

As you can see `Mage\Task\Factory` **is covered** but why are there other classes?
1. `Mage\Task\AbstractTask`  - because you create new tasks instances to test factory and they call abstract constructor. To prevent such behavior, we could add `@uses` annotation (as I did in #197) but we don't test class' constructors, so here `AbstractTask` coverage is underisable.
2. `Doctrine\Instantiator` - my eyes opened widely with big **WTF?!**. What is this class doing here? I think you ask yourself the same :)

Next thing I did was to add `@covers` annotations. Results weren't better. I've got only risky tests with exact places where those class were used.

So basically we can't add just `@uses` annotation because `TaskFactory` doesn't use anything from the outside. `@covers` is needed to mark test as risky since it tries to cover another class. We don't want to cover tasks' constructors. The only responsibility of this factory is to find out if the class exist and create new instantce of it.

eps90:
One more thing, that I haven't explain.
You can call me unit test purist but imagine you have a task, that does something with PHP extension you haven't got installed in task's constructor? Or with filesystem - it tries to get some directory contents. What if the directory doesn't exist? In both cases you got exceptions. The `Task\Factory` behaves correctly when testing manually but when you run your tests - they fail. 
Each test class should separate your testing class from **any** dependencies if you want to have **unit** tests. You can even move your class and its test into another project and when you run those test, they should pass.

SenseException:
I've created a helper function to mock every class instantiation by the factory as a test for creating more strict unit tests of the current code:

``` php
    private function createAlias($className)
    {
        $globalClassName = end(explode('\\', $className));

        $this->getMockBuilder('Mage\\Task\\AbstractTask')
            ->disableOriginalConstructor()
            ->setMockClassName($globalClassName)
            ->getMock();

        /**
         * current workaround
         * @link https://github.com/sebastianbergmann/phpunit-mock-objects/issues/134
         */
        if (!class_exists($className, false)) {
            class_alias($globalClassName, $className);
        }
    }
```

and used it inside the factory tests. e.g.

``` php
    public function testGet()
    {
        $this->createAlias('\\Mage\\Task\\BuiltIn\\Composer\\InstallTask');

        $task = Factory::get('composer/install', $this->config);
        $this->assertInstanceOf('\\Mage\\Task\\BuiltIn\\Composer\\InstallTask', $task);
    }
```

Since the original class InstallTask isn't loaded by autoloading, it will be replaced by the MockBuilder created mock. But the constructor is still called by it because it is set as final.

``` php
    public final function __construct(Config $config, $inRollback = false, $stage = null, $parameters = array())
    {
        throw new Exception('Exception will be called because constructor is final');
        $this->config = $config;
        $this->inRollback = $inRollback;
        $this->stage = $stage;
        $this->parameters = $parameters;
    }
``` 
The Exception was temporarly added for my test purpose.

It would be much easier when the Tasks have an interface that can be mocked. As best practice, the check inside the factory should be against an interface instead of the abstract class.

The final in constructor is used because all the current tasks are supposed to have the same method parameters. I'm sure this was made so the factory can always assume the same parameters for it for injecting.
It is technically possible to add a constructor to an interface, but I'm sure there are kitten dying somewhere because of such a coding idea.

Plus:
Keep the Exception in AbstractTask constructor but remove final and create the following test:

``` php
    public function testMocky()
    {
        $this->getMockBuilder('Mage\\Task\\AbstractTask')
            ->disableOriginalConstructor()
            ->setMockClassName('MyTask')
            ->getMock();

        $task = new \MyTask($this->config);

        $this->assertTrue(true);
    }
```

The Exception is still thrown. That may be because of the nature of `setMockClassName` and its instantiated class.

We can maybe fix the coverage with the current code, but I guess we need to refactor the factory or/and Task for more stricter purely tests.

The instantiator of doctrine is used by phpunit-mock-objects and only appears by creating the mock with disabled constructor. It is funny, that it appears in the coverage result, since the other libraries phpunit is using don't appear in it.

SenseException:
@eps90 I made up a solution for the Instantiator, since it isn't really used by the Factory but by the tests itself and shouldn't be a part for the uses-annotation. Please take a look.

eps90:
Looks ok for me, I'll take a closer look later, when I'll come back home :)

eps90:
Please accept my apologies again, I have completely no time recently to develop Magallenes. Merging! :)

SenseException:
No problem. We're doing this in our free time.

PR #178, opened 2015-01-09
switch the »current« symlink in an atomic way

Engerim:


andres-montanez:
Nice approach.
Thanks!

PR #181, opened 2015-01-14
fix bug in ApplyFaclsTask due to $cdToDirectoryFirst=true

KillerDiller:
Make `ApplyFaclsTask` compatible with default `AbstractTask::runCommandRemote()` parameter  `$cdToDirectoryFirst=true`.
When a relative path is given in `deployment -> to`, the task will first `cd` into this directory and then also prepend that path to the `setfacl` argument. One of these is enough. (This is true also for absolute paths given in `deployment -> to`, but will not case `setfacl` to fail.)

### DIFF
--- a/Mage/Task/BuiltIn/Filesystem/ApplyFaclsTask.php
+++ b/Mage/Task/BuiltIn/Filesystem/ApplyFaclsTask.php
@@ -25,7 +25,6 @@ class ApplyFaclsTask extends AbstractTask implements IsReleaseAware
     public function run()
     {
         $releasesDirectory = $this->getConfig()->release('directory', 'releases');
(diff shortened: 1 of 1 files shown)


eps90:
Good spot. :+1:

dasrick:
+1

eps90:
Ummm... by the way. You made a Pull Request to `master` branch. Since it's not a critical bugfix, please change it to be merged to `develop`.

KillerDiller:
@eps90, look again.

eps90:
Ahh, ok. Sorry but I was tired yesterday evening and I saw different thing that actually is. 
But it's quite strange that you did changes on `master` and you are merging it onto `develop`.

andres-montanez:
Thanks everyone!

PR #186, opened 2015-01-25
Ignoring errors while deleting temp files

d-ulyanov:
Release should not fail if temp files not found

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,7 @@ mage.phar
 ehthumbs.db
 Icon?
 Thumbs.db
+
+# IDE generated files
+.idea
 nbproject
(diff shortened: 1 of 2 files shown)


eps90:
What kind of errors were there?

eps90:
Okay, when I saw the PR, I was just-after wake up :) Sorry about the question above. 
Seems good, checking if file exist would only increase the complexity. So for me it's ok.

d-ulyanov:
Ok, could you merge it? :)

d-ulyanov:
We don't need to check that tmp file exists - just delete.

eps90:
I'll test that first (which seems to be ok) and then merge :) I'll do it ASAP.

eps90:
Seems everything is OK ;) I hadn't noticed any errors. Thanks a lot for this contribution! Merging! :)

andres-montanez:
Thanks everyone!!

Issue #188, opened 2015-02-02
Make Task-Factory more dynamic


Nilson89:
At the moment it is not possible to add new Tasks outside the .mage/task directory.

If you want to define Tasks to use in more than one project, you should be able to
1. Have a repository with composer support
2. Add this repository to projects dependencies in composer.json
3. Let the Task-Factory look for tasks like MyNamespace\Task\MyTask

chrisfromredfin:
#189

andres-montanez:
Hi, Version 1 of Magallanes has long reached it's end of life cycle, this feature will be available in the next mayor version.
Many thanks!

Issue #113, opened 2014-08-21
Use Composer autoloader for class loading to give more extendability for Magallan


eps90:
Hello again!
I've been mastering with Magallanes recently and I think it could be nice feature to use composer autoloader for Magallanes class loading. I know that the existing `Mage\Autoload` class is semi-convetional and takes base_dir dynamically, but there are many pros with going this way:
- optimized autoloader - and less work with supporting the current one
- common autoloading for all namespaces in the project - could be helpful with using utility libraries from project where mage is installed _locally_
- **ability to create external extensions repositories** - that's the thing I miss the most.
### _What's my idea about extensions?_

I use many projects based on Symfony2 and every time I'd need to copy my tasks to `.mage/tasks` directory. If I would create an external repo with tasks and commands, like `Symfony2MageExtensionsBundle` it would be much easier to implement and develop them.
### _How extensions should be implemented and developed?_

That's good question. Currently, Magallanes needs tasks and commands to be implemented in corresponding subdirectories in `.mage/` directory. We could create some DI-like mechanism or some index for tasks with full-named-class responsible for them, to get Magallan know, where to search for them:

```
# .mage/tasks.yml
tasks:
  symfony2/cache-clear: MyCustomBundle\MageTasks\SymfonyCacheClearTask
  doctrine/fixtures-load: MyCustomBundle\MageTasks\DoctrineFixturesLoadTask
commands:
  ...
# etc...
```

Of course each of these classes would implement corresponding Abstract class (which would be possible due to project-scoped autoload mechanism, yay!)
### _Ok, but isn't it more complicated than simple copying the tasks into my project?_

The idea above actually can look a bit complicated. But first advantage is to develop one task in one library, which makes everything easier out-of-the-box.
But let's say that some dependant library has implemented a hundred of tasks and for some reason we **must** to use them all at once, the thing becomes a bit problematic. The solution for this issue would be implementing an **extension class** like many extensionable libraries does. The class would parse its own `configuration.yml` file with tasks list and give the application ability to use it.

```
# Acme\MyCustomBundle\config\extension.yml
extension:
  tasks:
    symfony2/cache-clear: MyCustomBundle\MageTasks\SymfonyCacheClearTask
  commands:
    ...
```

Then loading an extension would be as simple as:

```
# .mage/extensions.yml
extensions:
  - Acme\MyCustomBundle\MyCustomExtension
```
### _Looks nice, what are the disadvantages?_

[...]

PR #189, opened 2015-02-02
Make Task Factory load tasks from external namespaces

Nilson89:
See https://github.com/andres-montanez/Magallanes/issues/188

### DIFF
--- a/bin/mage
+++ b/bin/mage
@@ -18,8 +18,8 @@ define('MAGALLANES_DIRECTORY', $baseDir);
 
 if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
     require_once __DIR__ . '/../vendor/autoload.php';
-} elseif (file_exists(__DIR__ . '/../autoload.php')) {
-    require_once __DIR__ . '/../autoload.php';
+} elseif (file_exists(__DIR__ . '/../../../autoload.php')) {
+    require_once __DIR__ . '/../../../autoload.php';
 } else {
     require_once $baseDir . '/Mage/Autoload.php';
     $loader = new \Mage\Autoload();
--- a/bin/mage
+++ b/bin/mage
@@ -18,6 +18,8 @@ define('MAGALLANES_DIRECTORY', $baseDir);
 
 if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
     require_once __DIR__ . '/../vendor/autoload.php';
+} elseif (file_exists(__DIR__ . '/../autoload.php')) {
+    require_once __DIR__ . '/../autoload.php';
 } else {
     require_once $baseDir . '/Mage/Autoload.php';
     $loader = new \Mage\Autoload();
--- a/tests/MageTest/Task/FactoryTest.php
+++ b/tests/MageTest/Task/FactoryTest.php
@@ -13,6 +13,9 @@ class FactoryTest extends PHPUnit_Framework_TestCase
 {
     public function setUp()
     {
+        $this->configMock = $this
+            ->getMock('Mage\Config');
+
         require_once __DIR__ . '/MyTaskExample.php';
     }
 
@@ -23,14 +26,28 @@ class FactoryTest extends PHPUnit_Framework_TestCase
     {
         $factory = new Factory;
 
-        $configMock = $this
-            ->getMockBuilder('Mage\Config')
-            ->getMock();
-
-        $myTask = $factory->get('My\Task\Example', $configMock);
+        $myTask = $factory->get('My\Task\Example', $this->configMock);
 
         $this->assertInstanceOf('My\Task\Example', $myTask);
 
+    }
+
+    /**
+     * @covers Mage\Task\Factory
+     */
+    public function testLoadClassWithoutNamespace()
+    {
+        $factory = new Factory();
+
+        $mock = $this
+            ->getMockBuilder('Mage\Task\AbstractTask')
+            ->setMockClassName('MyMageCustomTask')
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $task = $factory->get('MyMageCustomTask', $this->configMock);
+
+        $this->assertInstanceof('MyMageCustomTask', $task);
     }
 }
 
--- a/tests/MageTest/Task/FactoryTest.php
+++ b/tests/MageTest/Task/FactoryTest.php
@@ -9,13 +9,16 @@ use PHPUnit_Framework_TestCase;
  * @group Mage_Task
  * @coversDefaultClass Mage\Task\Factory
  */
-class ColorsTest extends PHPUnit_Framework_TestCase
+class FactoryTest extends PHPUnit_Framework_TestCase
 {
     public function setUp()
     {
         require_once __DIR__ . '/MyTaskExample.php';
     }
 
+    /**
+     * @covers Mage\Task\Factory
+     */
     public function testLoadFromExternalNamespace()
     {
         $factory = new Factory;
--- a/tests/MageTest/Task/FactoryTest.php
+++ b/tests/MageTest/Task/FactoryTest.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace MageTest\Task;
+
+use Mage\Task\Factory;
+use PHPUnit_Framework_TestCase;
+
+/**
+ * @group Mage_Task
+ * @coversDefaultClass Mage\Task\Factory
+ */
+class ColorsTest extends PHPUnit_Framework_TestCase
+{
+    public function setUp()
+    {
+        require_once __DIR__ . '/MyTaskExample.php';
+    }
+
+    public function testLoadFromExternalNamespace()
+    {
+        $factory = new Factory;
+
+        $configMock = $this
+            ->getMockBuilder('Mage\Config')
+            ->getMock();
+
+        $myTask = $factory->get('My\Task\Example', $configMock);
+
+        $this->assertInstanceOf('My\Task\Example', $myTask);
+
+    }
+}
+
--- a/Mage/Task/Factory.php
+++ b/Mage/Task/Factory.php
@@ -47,7 +47,7 @@ class Factory
         $taskName = str_replace(' ', '', $taskName);
 
         if (class_exists($taskName)) {
-            $className = $className;
+            $className = $taskName;
         } elseif (strpos($taskName, '/') === false) {
             $className = 'Task\\' . $taskName;
         } else {
--- a/Mage/Task/Factory.php
+++ b/Mage/Task/Factory.php
@@ -46,9 +46,10 @@ class Factory
         $taskName = ucwords(str_replace('-', ' ', $taskName));
         $taskName = str_replace(' ', '', $taskName);
 
-        if (strpos($taskName, '/') === false) {
+        if (class_exists($taskName)) {
+            $className = $className;
+        } elseif (strpos($taskName, '/') === false) {
             $className = 'Task\\' . $taskName;
-
         } else {
             $className = 'Mage\\Task\\BuiltIn\\' . str_replace(' ', '\\', ucwords(str_replace('/', ' ', $taskName))) . 'Task';
         }
--- a/docs/example-config/.mage/config/environment/staging.yml
+++ b/docs/example-config/.mage/config/environment/staging.yml
@@ -18,6 +18,7 @@ tasks:
 #    - sampleTask
 #    - failTask
     - scm/update
+    - My\Own\Namespace\Task
   on-deploy:
     - privileges
 #    - sampleTask
--- a/tests/MageTest/Task/MyTaskExample.php
+++ b/tests/MageTest/Task/MyTaskExample.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace My\Task;
+
+use Mage\Task\AbstractTask;
+
+class Example extends AbstractTask
+{
+    public function getName()
+    {
+        return 'I am an example task!';
+    }
+
+    public function run()
+    {
+        $command = 'echo "I do nothing"';
+        return $this->runCommandLocal($command);
+    }
+}


tests/MageTest/Task/FactoryTest.php
```diff
@@ -0,0 +1,33 @@
...
+namespace MageTest\Task;
+
+use Mage\Task\Factory;
+use PHPUnit_Framework_TestCase;
+
+/**
+ * @group Mage_Task
+ * @coversDefaultClass Mage\Task\Factory
+ */
+class ColorsTest extends PHPUnit_Framework_TestCase
```
eps90: Hey, copy-paster! ;p I think it isn't `ColorsTest` class, isn't it ? ;)

tests/MageTest/Task/FactoryTest.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    public function setUp()
+    {
+        require_once __DIR__ . '/MyTaskExample.php';
+    }
+
+    public function testLoadFromExternalNamespace()
```
eps90: Could you add coverage annotation?
```php
/**
 * @covers ::get
 */
```

eps90:
Please provide also some examples in `docs` directory.

Nilson89:
Changed naming of FactoryTest and added example on how to run Task of own namespace

eps90:
Looks a lot better. So, everyone, be free to test it ;) Looks like a really nice feature :)

tests/MageTest/Task/FactoryTest.php
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * @covers Mage\Task\Factory
+     */
+    public function testLoadFromExternalNamespace()
+    {
+        $factory = new Factory;
+
+        $configMock = $this
+            ->getMockBuilder('Mage\Config')
```
eps90: You don't need to instantionate `MockBuider` here. It'll be ok to do `$this->getMock('Mage\Config')`

tests/MageTest/Task/FactoryTest.php
```diff
@@ -0,0 +1,36 @@
...
+            ->getMock();
+
+        $myTask = $factory->get('My\Task\Example', $configMock);
```
eps90: I'm interested in one thing. Since tasks can be defined with namespace, so please test the following:
* Remove `MyTaskExample` class
* Create a mock of this task, but without namespace (PHPUnit's bug doesn't allow to name classes with namespaces):
```php
$this->getMockBuilder('Mage\Command\AbstractCommand')
    ->setMockClassName('MyTaskExample')
    ->getMock();
```
* Try to get your class with factory by:
```php
$myTask = $factory->get('MyTaskExample');
```
* and finally assert:
```php
$this->assertInstanceof('MyTaslExample', $myTask);
```

Please try this thing and tell me how it went :) Thanks!

Nilson89:
Hey, what´s the status of this PR?

eps90:
Sorry, I though anyone else wants to test it.
By the way, you can correct one more thing. You have redudant class that you added only for testing. Look at #167, you can do it on more clean way :)

bin/mage
```diff
@@ -18,6 +18,8 @@ define('MAGALLANES_DIRECTORY', $baseDir);
 
 if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
     require_once __DIR__ . '/../vendor/autoload.php';
+} elseif (file_exists(__DIR__ . '/../../../autoload.php')) {
+    require_once __DIR__ . '/../../../autoload.php';
```
SenseException: Where does this actually lead to?

eps90: @SenseException It's partly our fault, because these lines are in `master` branch but has not been  merged into `develop`. See 4cce76472e6ee67d6a88d1abe36c39f403aca7a1
I'll try to merge it locally and see if it won't crash anything (it shouldn't though)

SenseException:
This naming of the tasks in config is more explaining then the current ones, but it is inconsistent. I'm wondering if we should implement a configurable namespace autoloading for task-classes. Maybe something like Doctrine uses for annotations. It would look like this `TaskRegistry::registerAutoloadNamespace("My\Own\Namespace", "/path/to/my/own/namespace-files");`.

As a Fallback, the custom task and the build-in task namespaces are added by default. We could eventually ditch the hardcoded namespaces and if-else blocks from the factory.

eps90:
@SenseException few months ago I've proposed some idea about extensions mechanism in Symfony2.  You can look on it at #113.  I've got something implemented but I've stopped for a moment to make sure that everything is unit tested so from that moment we have a rush with writing unit test. This PR introduces a "tracer bullet" for the final idea solution. So I think it's a good step to create communication with classes out of Magallanes core files.

SenseException:
@eps90 We would be able to create tasks in an own repository/project. My idea is, that we wouldn't need to use the whole namespace but the naming like we are used to in Magallanes. Extern task classes, not in .mage/task or in the build-in directory, but still with namings like:

``` yml
tasks:
    pre-deploy:
        - my-task/do-pre-deploy-stuff
```

eps90:
@SenseException Of couse and your idea covers mine :) If you'd look at #113 you'll see that I use extensions mechanism, similar to Symfony2 one or Behat (which are bundles too).
What I want to say is that PR by @Nilson89 introduces first, working part of such extension. Of course it's not the way we would like to have it in future, but it works, I think. So here I can see two solutions:
* accept this merge request and keep combatibility in future solutions
* reject it and implement my solution
* others?

SenseException:
@eps90 I like @Nilson89 PR since I had wished to be able to use tasks from other namespaces from my first use of Magallanes, but I think we shouldn't introduce `My\Own\Namespace\MyTask` as possible task "name".

A solution (or maybe you like the static variant) could look like this since it is not an autoloading problem like I thought before:

Registering tasks
``` php
// register task, prefix and namespace come from config
$taskRegistry->registerNamespace('prefix', 'My\\Own\\Namespace');
```

Get the task class name in task factory
``` php
// in the factory, prefix/my-task is a task from the env yaml
$className = $taskRegistry->createTaskClass('prefix/my-task');
```

$taskRegistry would be nothing more than a container for key values. 

There would be the following benefits for this solution:

* We don't introduce anther task name convention, next to my-task and symfony2/cache-clear
* Removing hardcoded namespaces from task factory
** directories / namespaces in Mage\Task\BuiltIn\* and Task namespace are registered by default
* changing the name my-task to class name MyTask would happen in this registry class
* We don't need class_alias from #167 anymore (same with #176)
* Factory class gets smaller and easier to test, since we don't need if-elseif-else block anymore
* Depending how this registry class looks like, the same class may be also usable for Command Factory later
** no duplicated code of class naming (like my-task to MyTask is)
** no duplicated code for class name creation (removing hardcoded namespaces from command factory)

So I'm more for adding something like this. It is a pretty good addition for this PR and the Magallanes project itself. :-)

eps90:
@SenseException Of course I did understood your idea earlier and I like it a lot :) I also think that we should make the naming conventional for shared and built-in tasks. I can see your inspiration by Doctrine Annotations, but let me point few things (of course to consider, I've noticed them in the moment). If the text I gonna write covers your idea, then treat it like a solution proposition :)

Let's say Magallanes have your solution implemented. In meantime many bundles with Magallanes Tasks were created. So you have your main project, let's say `MyAwesomeProject` and there are projects with tasks:
* `MageSymfonyTasks` containing:
  * `Symfony\Cache` containing:
    * `CacheClearTask`
    * `CacheWarmupTask`
  * `Symfony\ProjectCleaner` containing:
    * `ClearEnvsTask`
    * `RecreateParametersTask`
* `MageDoctrineTasks` containig:
    * `Doctrine\Database` containing:
      * `SchemaUpdateTask`
      * `DatabaseCreateTask`
      * [...]
    * `Doctrine\Fixtures` containing:
      * `FixturesLoadTask`

and so on. I just wanted to show you the possible structure of possible plugins-container-projects.
Now, Let's configure them in target project:

```php
$taskRegistry->registerNamespace('cache', 'MageSymfonyTasks\Symfony\Cache');
$taskRegistry->registerNamespace('project-clean', 'MageSymfonyTasks\Symfony\ProjectCleaner');
$taskRegistry->registerNamespace('database', 'MageDoctrineTasks\Doctrine\Database');
$taskRegistry->registerNamespace('fixtures', 'MageDoctrineTasks\Doctrine\Fixtures');
```

What you need here is to know plugin project structure and register each namespace by yourself. Developer who'd need to configure 10 plugin projects, each containing 10 different namespaces wants to commit suicide. Moreover, he has to do it in every project he has.

----------

Now let me show you how this could be done following my idea.
Author of tasks bundle creates a configuration file / extension class / whatever else, showing available tasks for his project. E.g.:
```php
namespace MageSymfonyTasks;

class Extension implements ExtensionInterface
{
	public function getName()
	{
		return 'SymfonyTasks';
	}
	
	public function registerTasks()
	{
		$this->registerTaskNamespace('Symfony\Cache');
		// or even:
		$this->registerTask('cache-clear', 'Symfony\Cache\CacheClearTask');
	}
}
```
You find this project and decide to use it. Everything you need to do is register this extension in your target project:
```yaml
#some/extension/config/path.yml
extensions:
	- MageSymfonyTasks\Extension
	# or just:
	- MageSymfonyTasks
```
and you're done. Tasks contained in your extensions can be fetch by command:
```
bin/mage tasks:show
```
or by documentation shared by plugin's author.

----------
Also, creating `registerTaskNamespace` method would not prevent us from using `class_alias` in tests for searching for a class that does not exist because it's still a factory - you need to make sure that mock is registered, even if the class doesn't exist.

SenseException:
##### TaskRegistry

``` php
$taskRegistry->registerNamespace('cache', 'MageSymfonyTasks\Symfony\Cache');
$taskRegistry->registerNamespace('project-clean', 'MageSymfonyTasks\Symfony\ProjectCleaner');
$taskRegistry->registerNamespace('database', 'MageDoctrineTasks\Doctrine\Database');
$taskRegistry->registerNamespace('fixtures', 'MageDoctrineTasks\Doctrine\Fixtures');
```
This code is actually never created by the dev. The dev only uses the configuration of Magallanes. In my first post, I already mentioned that this would be configurable and I had something similar like your example in mind. But maybe weaker and simpler like key values.

``` yml
tasks:
    prefix: My\Own\Namespace
    more: My\Other\Namespace
   # not the build-in ones because they are already available
```

I like the idea of the extension for the case of multiple namespace paths like we know from the build-in tasks. Even though I don't think that there will never be "hundreds" of tasks in different needed in a project, the possibility of projects on github offering many tasks could make extensions come quite handy. The (possible) build-in extension of Magallanes contains the known build-in task namespaces and the one of .mage/tasks too.

I'm still thinking in smaller steps for improving the current code and the extensions are a bigger step, so if there is really a use case for it and instead of YAGNI it is YGNI, then it's going to be a cool feature.

##### class_alias in tests

Test code of #167:

``` php
$mock = $this->getMockBuilder('MyAbstractClass')
    ->setMockClassName('MyClass')
    ->getMock();
class_alias('MyClass', 'MyNamespace\MyClass');
```

`setMockClassName` can not be used for full namespaces, only for class name like `MyClass`. Since we aren't bound to a specific namespace anymore with this PR, the mocked `$taskRegistry->createTaskClass('my-class');` (or call it whatever you like) method can also return a class name of the global namespace. With this, `setMockClassName`of the MockBuilder can be used without a class_alias workaround.

"Which namespace" isn't relevant for the factory anymore. It only checks if the returned class name exists and if the created object is `AbstractTask` or an interface. The check about the type is another story. :-)

eps90:
Hmm ok, what exactly does `createTaskClass` do?

eps90:
@SenseException ping :)

SenseException:
Sorry, I'm currently short of time this week. I try to see if I can go back to it this evening.

eps90:
@SenseException sure, don't worry. I thought you've just forgot about that ;)

SenseException:
@eps90 The idea with the Registry came to me after I thought about the tests with class_alias and the nearly equal code in the Command Factory. Both are using hardcoded namespaces and we already know from sebastianbergmann/phpunit-mock-objects#134, we don't have to expect any changes from PHPUnit and I kind of understand that, even if it means that it is harder to test legacy code.

So I came up with a refactoring idea, which could be combined with this PR. The fictive `createTaskClass('prefix/task-name')` method combines the functionality of task name transformation

``` php
        $taskName = ucwords(str_replace('-', ' ', $taskName));
        $taskName = str_replace(' ', '', $taskName);
```

and creating the full class name

``` php
        if (class_exists($taskName)) {
            $className = $taskName;
        } elseif (strpos($taskName, '/') === false) {
            $className = 'Task\\' . $taskName;
        } else {
            $className = 'Mage\\Task\\BuiltIn\\' . str_replace(' ', '\\', ucwords(str_replace('/', ' ', $taskName))) . 'Task';
        }
```

We know that those hardcoded namespaces are a burden right now, which is why I thought, that the full class name could come from a method, that can be mocked and return a class name in the global namespace for the tests of the factory. This way, we don't have to bother with a class alias, since the factory doesn't care anymore where the class name belongs to. Because we still need to refer to the current build-in tasks and the custom tasks in .mage directory, I thought about registering the tasks like in the previous comments mentioned, because the build-in tasks are using this kind of naming already. This way, we don't have to maintain the if-else, and with this PR if-elseif-else, anymore. With the possibility to use this new class for Commands too. Of course it wouldn't be a TaskRegistry class for both but maybe a simple Registry one.

When I saw this PR, I was on one side pretty happy about that it doesn't matter in which namespace a task has to be stored now, but not satisfied by the thought that another totally different naming convention of tasks were introduced to the configuration and that the if-construct got more complex, which is more work for tests and kind of isn't really "boy scoutish".

While others introduce some great new features, like this PR, I'm more interested into the current code, that has to be maintained. In projects with no tests for the current code, I prefer doing small and lucid steps of changes. Extensions are a completly new Feature.

eps90:
Ok, so as far as I understood you you want to:
* create a container (you call it registry here)
* when the application starts, parse `tasks.yml` or `commands.yml` or whatever else and store it like:
```php
$registry = [
    'tasks' => [
        'task-one' => 'Mage\Tasks\FirstTask',
        'task-two' => 'Mage\Tasks\SecondTask'
        // ...
    ]
];
```
* in `Factory` - when there's a need to instantionate a task/command - create a task with given alias

If so, I like it very much ;) But I'm still not sure this solution overcomes the `class_alias` problem. In some `tasks.yml` parser class you need to check if the class exists. Otherwise, a user will see the an ugly notice that you try to create a class' object that doesn't exist. In tests we need to have some class mocks to test both cases. 
Still if I understood you well, if you register only namespaces to look for task/command classes, `Registry` class or another responsible for that could scan given namespaces for classes implementing given interface. It wouldn't be super-fast solution, but it also works as well. But if we want to create fictional classes that can exist or not, we can't overcome the problem with `class_alias`, because we need to check it. 
Please, maybe I don't understand you well, is that what I've written explains your idea?

SenseException:
Yes, something like that, but with namespaces.

``` php
$registry = [
    'tasks' => [
        'task-foo' => 'Mage\Tasks',
        // ...
    ]
];
```

works with FirstTask and SecondTask, because they are part of it: `task-foo/first-task`

##### Test

And the global namespace can also be registered, which is why this test works:

``` php
    public function testMocky()
    {
        $this->getMockBuilder('Mage\\Task\\AbstractTask')
            ->disableOriginalConstructor()
            ->setMockClassName('MyTask')
            ->getMock();

        $task = new \MyTask($this->config);

        $this->assertInstanceOf('Mage\\Task\\AbstractTask', $task);
    }
```

HugoCardenas:
Hi,
just wanted to ask what's the status of this.
I bumped into the same issue which this PR would fix, so if this is going forward, I won't need to fix it myself.
Thanks!

eps90:
@HugoCardenas Yes, discussuion has broken in a moment. I think we should create some mechanism with extendable registry (as me and @SenseException suggested) and then we can think about including external classes. This thing is needed and we don't want to change solution when final product will be ready. So, if you don't mind @Nilson89 , I think we should close this PR and the related issue and think about more advanced task loading mechanism.

chrisfromredfin:
I wish I knew more Symfony/OO! I would love to help, but not sure how. If this PR is closed, though - where is the issue/PR with the newer/more-current method of thinking, or does it not exist yet?

eps90:
@chrisfromredfin It depends :) One issue created by has been created once, but it's closed as the first step (composer autoloading) has been done. Discussion moved into this issue, so yes, I think we should create another task. I'll create one in spare time to include conclusions that already has been made here.

Ideally, we don't want to use Symfony here, since it's a tool with no dependencies (besides dev dependencies like `phpunit`). We need to create something extendable, reusable and well designed to make it easer to work with in the future. So we need to discuss about a little to find out some mockups and prototypes how such solution should work.

pfastre:
We're having hundreds of projects, each with their own .mage directory and with custom tasks. Right now we copy the tasks into each .mage/tasks directory with a shell script. So this feature would really be useful.
I understand a decision hasn't been made yet. Is this a question of disagreeing or of having no time to pick this up and make it into a priority? Is there a clear conclusion of how this should be implemented?

chrisfromredfin:
I think it's more that there's no "clear conclusion of how this should be
implemented" - I could say that if I could get pointed in a direction, I
may be able to pick up the torch, but I'm new to contributing to the
project so would need a little handholding. :)


--
Christopher J. Wells
Redfin Solutions, LLC
office: 908-4-DRUPAL (908-437-8725)
extension: 601
fax: 815.642.4600
chris@redfinsolutions.com
www.redfinsolutions.com

andres-montanez:
Many thanks to all!

PR #194, opened 2015-02-08
Added niceness for aggressive symfony commands

ramunasd:


eps90:
Shouldn't it be configurable?

andres-montanez:
Hi! I like the idea, but it should be configurable, as @eps90 recommended.
Perhaps I want to run it with a high priority in some case.

ramunasd:
Guys, how you gonna write tests for tasks? There is some wrong implementations that restricts from checking output/moking tasks.

eps90:
@ramunasd Unit testing coverage in the project is currently in progress. I know there are many anti-patterns but we can't do the refactoring until we cover the main parts of the application. I think you're thinking about final methods for running commands. Don't worry I have some local changes and I'm going to push something soon.

ramunasd:
@eps90 What do you mean ".. main parts of the application" ?

eps90:
Ideally we need to cover **all** the production code. But by writing that I meant critical points of code. It's first necessary step to **begin** refactor in my opinion. 
In the case of tasks, we need to cover `AbstractTask` and `Task\Factory` first and as I said before, I have something written already. As I said, there are many anti-patterns here, which makes mocking very difficult.

eps90:
Maybe you've got something more to add? Every note is valueable for us :)

ramunasd:
Thats all for now :)

eps90:
Great :) If you've got **any** issue about Magallanes, please report it immediately. Any feedback about project quality if helpful. And any of us - developers - agrees that we would like to have powerful deployment tool for PHP applications :)

Mage/Task/BuiltIn/Symfony2/SymfonyAbstractTask.php
```diff
@@ -26,7 +29,13 @@ protected function getAppPath()
         } else {
             $defaultAppPath = 'app/console';
         }
-
-        return $this->getConfig()->general('symfony_app_path', $defaultAppPath);
+        
+        $path = '';
+        $niceness = intval($this->getParameter('niceness', 0));
```
eps90: The parameter can be boolean value. We use Symfony's YAML component to parse configs' content, and it does support boolean values, as well as arrays, objects and `null`s.
So here you should write:
```php
$niceness = $this->getParameter('niceness', false);
if ($niceness) {
// ...
```

eps90: By the way, such thing shouldn't be done in `getAppPath` method, because you expect to get absolute or relative path to Symfony's `command` script. In the other hand there is no method for calling explicitly concrete Symfony2 command .

ramunasd:
You are right, niceness should be as task runner option, independent from the task

PR #195, opened 2015-02-09
Fix 1st deploy bug

spongeben:
Fix 2 problems leading to a 1st deploy failure (one only with rsync strategy and the other more widely I think)

In `ReleaseTask.php`, we're trying to change owner of symlink file. But on first release the symlink does not exists yet so the release fail (problem comes from #178 I think). I've updated to only check owner if symlink exists. (maybe we can remove totaly the symlink owner check as it's created after ?)

In `RsyncTask.php`, with `rsync_copy` enabled, if the previous release folder does not exists, the release directory is not created, and rsync fail. I've updated to revert to manual empty directory creation if there is no previous release.

### DIFF
--- a/Mage/Task/BuiltIn/Deployment/ReleaseTask.php
+++ b/Mage/Task/BuiltIn/Deployment/ReleaseTask.php
@@ -74,12 +74,13 @@ class ReleaseTask extends AbstractTask implements IsReleaseAware, SkipOnOverride
                 }
             }
 
-            if ($resultFetch && $userGroup != '') {  
-                $command = 'chown -h ' . $userGroup . ' ' . $symlink
-                    . ' && '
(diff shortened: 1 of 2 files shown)


eps90:
Whoa, the code here is so complecated that it's almost unreadable. Surely the feedback is needed here but for me this should work.

robi50:
+1

It should merged. aab71eceac19697f33fea8abf73f19b8eb0e84cf is closes #199  (which is definitely an important bug).

eps90:
Hey guys. Sorry but I didn't know how serious bug it is until today. Merging immediately!
@robi50 I think I can close your issue and PR since this one, containg a bit more, fixes one more bug in RSync. Thanks for all!

Issue #163, opened 2014-12-02
Adding Coverage to Magallanes code


SenseException:
Since the first tests are in development branch, we should think about adding code coverage to the project.

I saw that some projects are using Coveralls (https://coveralls.io/), but I'm not that familiar with this service. Are there other code coverage services that can be added to github?

eps90:
@SenseException @jhuet Yup, we should add code coverage indicator.
Developers should also run tests with code coverage reporting, e.g. by:

    ./bin/phpunit --coverage-text

or even

    ./bin/phpunit --coverge-html report

PS1: The `bin/phpunit` has been added by in #162 . It still waits for approval...
PS2: @jhuet As you've mentioned in #174 , we need to compose some documentation for contributors. With no rules here's gonne be the armageddon...

eps90:
@SenseException @jhuet I've added a PR with travis and coveralls configuration. Please take a look on it.

eps90:
Issue moved to #201 . Closing.

PR #196, opened 2015-02-14
Add Travis and Coveralls support

eps90:
Hi all!
This PR introduces **travis-ci** and **coveralls.io** support as it has been proposed in #163.
Actually there was a config for travis, but I've modified it by:

* removing `--dev` flag in `composer install` since it's the default option when running `composer install`
* add scripts necessary to trigger coveralls report

I needed to add `php-coveralls` package to composer to run Coveralls reporting. I've also add `.coveralls.yml` to override default options of coveralls build.

**Since I'm not the owner of this repo, @andres-montanez, please turn it on in repo management panel on travis-ci.org and coveralls.io before or just after merging this PR**. I've tested both on my fork and everything build with success.

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,2 @@
+src_dir: Mage
+coverage_clover: build/logs/coverage.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,10 @@ php:
 
 install:
   - composer install --prefer-source
+
+script:
+  - mkdir -p build/logs
+  - bin/phpunit --coverage-clover build/logs/coverage.xml
+
+after_script:
+  - bin/coveralls -v
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ php:
   - 5.6
 
 install:
+  - composer self-update
   - composer install --prefer-source
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,4 @@ php:
   - 5.6
 
 install:
-  - composer install --dev --prefer-source
+  - composer install --prefer-source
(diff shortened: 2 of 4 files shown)


SenseException:
@eps90 Good to see, that you added the composer self-update. I missed that part at the first time. Did you put the coverage generation to travis config instead phpunit.xml, so the coverage isn't build on a dev's machine by executing phpunit?

@andres-montanez How about adding the travis build and the coverall image into the readme markdown file? I remember I've add the build one to it before, but of course without activating it for your repo, it won't work. Would be nice to have these two there.

eps90:
@SenseException Yes, I didn't change the configuration, because some of some reasons / rules that I follow in my everyday work:
* developers want to execute their tests almost in real time, while coverage generation, especially in big project, may slow down that process
* connecting to the point above, developer should decide by himself what coverage report he wants to use. For local environment, I think, the best way is to generate HTML/text output, while for CI build, XML one. The configuration would generate all redudant reports which is not needed.

But, as always, I'd like to hear more about some *best practices* for running coverage. Maybe the way you proposed is better, but so far I used "my way" and no one complains :)

So, if a developer wants to run tests with coverage, he should do it manually. Anyways, travis should warn us if coverage levels falls.

I thought about adding also a dummy image to readme file, but I decided it should be done in separate pull request or my @andres-montanez himself to avoid inconsistency, while travis and coveralls are not turned on.

SenseException:
@eps90 I think this PR is okay.

eps90:
@SenseException Well I'm sure :) We can merge it but travis and coveralls won't work if @andres-montanez won't enable it in project settings. I can merge it and open new issue about **enabling** it, what do you think?

SenseException:
@eps90 I agree.

eps90:
You're right. I'm merging it then. Issue moved to #201 .

PR #197, opened 2015-02-22
Add commands tests

eps90:
Hey there!
This PR introduces tests of few commands including:
* `AbstractCommand`
* `CompileCommand`
* `ListCommand`
* `LockCommand`
* `UnlockCommand`

These tests cover the code in 100% (line coverage) without changing current functionality.
Few words about PR:
* I used [malkusch\php-mock](https://github.com/malkusch/php-mock) for creating mocks of PHP's built-in functions. As far as I know, it uses the fact that PHP interpreter will use the function with the same name from the namespace where it was used in. This library contains some other cool options like making **stubs** on built-in functions but since we're keeping compatibility with PHP 5.3, we can't use them and we can just mock them to avoid unexpected results. Fortunatelly, author created the `php-5.3` branch.
* I had to use `@uses` annotation to point that we use some other classes (like statics, impossible to inject and then mock) and not cover them. Without that annotation PHPUnit run with `--coverage-*` trigger will mark all tests as risky. **Note:** I didn't use this annotation only to suppress warnings. Each test class was tested without them to see if there are no other classes under control. As you can see, each test class contains almost the same `@uses` annotations.
* I had to add setter in `CompileCommand` to mock `Mage\Compiler` class. This setter is called in (also created by me) constructor, so nothing changed in this class' functionality.

For now coverage of the project is **22.36%**. Before creating these tests, it was about 7%.
Feel free to review and discuss! Thanks!

### DIFF
--- a/Mage/Command/BuiltIn/CompileCommand.php
+++ b/Mage/Command/BuiltIn/CompileCommand.php
@@ -22,6 +22,21 @@ use Mage\Compiler;
 class CompileCommand extends AbstractCommand
 {
     /**
+     * @var Compiler
+     */
+    private $compiler;
+
+    public function __construct()
+    {
+        $this->compiler = new Compiler();
+    }
+
+    public function setCompiler(Compiler $compiler)
+    {
+        $this->compiler = $compiler;
+    }
+
+    /**
      * @see \Mage\Compile::compile()
      */
     public function run()
--- a/Mage/Command/BuiltIn/CompileCommand.php
+++ b/Mage/Command/BuiltIn/CompileCommand.php
@@ -42,13 +42,23 @@ class CompileCommand extends AbstractCommand
     public function run()
     {
         if (ini_get('phar.readonly')) {
-            Console::output('The <purple>php.ini</purple> variable <light_red>phar.readonly</light_red> must be <yellow>Off</yellow>.', 1, 2);
+            Console::output(
+                'The <purple>php.ini</purple> variable <light_red>phar.readonly</light_red>'
+                . ' must be <yellow>Off</yellow>.',
+                1,
+                2
+            );
+
             return 200;
         }
 
         $this->compiler->compile();
 
-        Console::output('<light_purple>mage.phar</light_purple> compiled <light_green>successfully</light_green>', 0, 2);
+        Console::output(
+            '<light_purple>mage.phar</light_purple> compiled <light_green>successfully</light_green>',
+            0,
+            2
+        );
 
         return 0;
     }
--- a/Mage/Command/BuiltIn/CompileCommand.php
+++ b/Mage/Command/BuiltIn/CompileCommand.php
@@ -28,7 +28,7 @@ class CompileCommand extends AbstractCommand
 
     public function __construct()
     {
-        $this->compiler = new Compiler();
+        $this->setCompiler(new Compiler());
     }
 
     public function setCompiler(Compiler $compiler)
@@ -46,8 +46,7 @@ class CompileCommand extends AbstractCommand
             return 200;
         }
 
-        $compiler = new Compiler;
-        $compiler->compile();
+        $this->compiler->compile();
 
         Console::output('<light_purple>mage.phar</light_purple> compiled <light_green>successfully</light_green>', 0, 2);
 
--- a/Mage/Command/BuiltIn/CompileCommand.php
+++ b/Mage/Command/BuiltIn/CompileCommand.php
@@ -26,13 +26,12 @@ class CompileCommand extends AbstractCommand
      */
     private $compiler;
 
-    public function __construct()
+    public function __construct(Compiler $compiler = null)
     {
-        $this->setCompiler(new Compiler());
-    }
+        if ($compiler === null) {
+            $compiler = new Compiler();
+        }
 
-    public function setCompiler(Compiler $compiler)
-    {
         $this->compiler = $compiler;
     }
 
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "4.3.5",
-        "satooshi/php-coveralls": ">=0.6.1"
+        "satooshi/php-coveralls": ">=0.6.1",
+        "malkusch/php-mock": "dev-php-5.3"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
         "psr-4": {
             "Mage\\": "./Mage",
             "Task\\": [".mage/tasks", "../../../.mage/tasks"],
-            "Command\\": [".mage/tasks", "../../../.mage/commands"]
+            "Command\\": [".mage/tasks", "../../../.mage/commands"],
+            "MageTest\\": "./tests/MageTest"
         }
     },
     "config": {
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,11 @@
         "psr-4": {
             "Mage\\": "./Mage",
             "Task\\": [".mage/tasks", "../../../.mage/tasks"],
-            "Command\\": [".mage/tasks", "../../../.mage/commands"],
+            "Command\\": [".mage/tasks", "../../../.mage/commands"]
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
             "MageTest\\": "./tests/MageTest"
         }
     },
--- a/tests/MageTest/Command/AbstractCommandTest.php
+++ b/tests/MageTest/Command/AbstractCommandTest.php
@@ -8,6 +8,7 @@ use PHPUnit_Framework_MockObject_MockObject;
 /**
  * Class AbstractCommandTest
  * @package MageTest\Command
+ * @author Jakub Turek <ja@kubaturek.pl>
  * @coversDefaultClass Mage\Command\AbstractCommand
  */
 class AbstractCommandTest extends BaseTest
--- a/tests/MageTest/Command/AbstractCommandTest.php
+++ b/tests/MageTest/Command/AbstractCommandTest.php
@@ -2,6 +2,7 @@
 
 namespace MageTest\Command;
 use Mage\Command\AbstractCommand;
+use MageTest\TestHelper\BaseTest;
 use PHPUnit_Framework_MockObject_MockObject;
 
 /**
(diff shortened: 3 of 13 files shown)


SenseException:
If I can make it possible, I'll take a look at it over the weekend. It is too bad that Reflection is needed, but maybe we'll get some inspiration for other ideas for this ticket and #175 sometime.

eps90:
@SenseException Yes, I know it's ugly that I used Reflection. This is the way I test value objects. I didn't want to set it by setting by setter and get by getter. I've never found any other way to do this.
In the other hand, I could omit testing those methods since their complexity is 1.. and I do simple assignment there. So I can omit them or disable coverage there ;) Especially it's just an abstract class.
Thanks for feedback! ;)

Mage/Command/BuiltIn/CompileCommand.php
```diff
@@ -22,19 +22,43 @@
...
     /**
+     * @var Compiler
+     */
+    private $compiler;
+
+    public function __construct()
+    {
+        $this->setCompiler(new Compiler());
+    }
+
```
SenseException: Why not the following constructor code?

``` php
    /**
     * @var Compiler
     */
    private $compiler;

    public function __construct(Compiler $compiler = null)
    {
        if (is_null($compiler)) {
            $compiler = new Compiler();
        }
        
        $this->compiler = $compiler;
    }
```

This way, you don't need the setter and won't allow to change the compiler on runtime like the previews version.

Mage/Command/BuiltIn/CompileCommand.php
```diff
@@ -22,19 +22,43 @@
...
+    public function setCompiler(Compiler $compiler)
```
SenseException: If you want to keep the setter and the possibility to set the compiler on runtime, the doc-block is missing for this method.

composer.json
```diff
@@ -9,13 +9,15 @@
...
+        "phpunit/phpunit": "4.3.5",
+        "malkusch/php-mock": "dev-php-5.3"
     },
     "autoload": {
         "psr-4": {
             "Mage\\": "./Mage",
             "Task\\": [".mage/tasks", "../../../.mage/tasks"],
-            "Command\\": [".mage/tasks", "../../../.mage/commands"]
+            "Command\\": [".mage/tasks", "../../../.mage/commands"],
+            "MageTest\\": "./tests/MageTest"
```
SenseException: For unit test classes, please use autoload-dev.
https://getcomposer.org/doc/04-schema.md#autoload-dev

tests/MageTest/Command/AbstractCommandTest.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->abstractCommand->setConfig($configMock);
+
+        $actual = $this->getPropertyValue($this->abstractCommand, 'config');
+        $this->assertEquals($configMock, $actual);
+    }
+
+    /**
+     * @covers ::getConfig
+     */
+    public function testGetConfig()
```
SenseException: It is considered for some people as not a good practice, but what do you think about the depends-annotation? Maybe not a separated unit running for itself, but a possibility not to use Reflection for the getter test.

tests/MageTest/Command/AbstractCommandTest.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * @covers ::setConfig
+     */
+    public function testSetConfig()
+    {
+        $configMock = $this->getMock('Mage\Config');
+        $this->abstractCommand->setConfig($configMock);
+
+        $actual = $this->getPropertyValue($this->abstractCommand, 'config');
+        $this->assertEquals($configMock, $actual);
```
SenseException: You actually expect the same instance you set with the setter, not another config instance. You need to use

``` php
$this->assertSame($configMock, $actual);
```

tests/MageTest/Command/AbstractCommandTest.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * @covers ::getConfig
+     */
+    public function testGetConfig()
+    {
+        $configMock = $this->getMock('Mage\Config');
+        $this->setPropertyValue($this->abstractCommand, 'config', $configMock);
+
+        $actual = $this->abstractCommand->getConfig();
+        $this->assertEquals($configMock, $actual);
```
SenseException: Same as mentioned in your setter test. You need `assertSame`:

``` php
$this->assertSame($configMock, $actual);
```

tests/MageTest/Command/BuiltIn/CompileCommandTest.php
```diff
@@ -0,0 +1,114 @@
...
+        $iniGetMock->enable();
+
+        $this->setUpConsoleStatics();
+    }
+
+    /**
+     * @covers ::__construct
+     * @covers ::setCompiler
+     */
+    public function testConstruct()
```
SenseException: Does a constructor needs an own test when it only sets a property?

tests/MageTest/Command/BuiltIn/CompileCommandTest.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        $compilerProperty = $this->getPropertyValue($this->compileCommand, 'compiler');
+        $this->assertInstanceOf('Mage\Compiler', $compilerProperty);
+    }
+
+    /**
+     * @covers ::__construct
+     * @covers ::setCompiler
+     */
+    public function testSetCompiler()
```
SenseException: See my comment in Compiler class part about Setter.

composer.json
eps90: good point.

Mage/Command/BuiltIn/CompileCommand.php
eps90: Why?

tests/MageTest/Command/AbstractCommandTest.php
eps90: I agree, I wanted just test the current code with almost no changes. I also agree that dependency injection would help us a lot but for now I wanted only to test if the classes are constructed properly, nothing else. i think we shouldn't do any refactor till we test the weakest points of the project.

eps90:
Thanks @SenseException . I look at your comments in free time. Now I'm quite busy, but I think tomorrow evening I'll do some fixes into my PR. Thanks a lot! ;)

tests/MageTest/Command/BuiltIn/UnlockCommandTest.php
```diff
@@ -0,0 +1,125 @@
...
+        self::$isFileExists = false;
+
+        $actualExitCode = $this->unlockCommand->run();
+        $expectedExitCode = 0;
+
+        $this->assertFalse(self::$isUnlinkCalled);
+        $this->assertEquals($expectedExitCode, $actualExitCode);
+        $this->assertEquals($expectedLockFilePath, self::$fileExistsResult);
+    }
+}
```
SenseException: Both tests,testRun and testRunIfFileNotExist, are sharing nearly the same code. Maybe this can be excluded into a common used method.

eps90: cetrainly :)

Mage/Command/BuiltIn/CompileCommand.php
SenseException: For the different IDE's that are using it. Some may need it and it may help other contributors.

And maybe someday we create a API documentation for the project :-)

eps90: API is ready but it still need improvements. It should be updated per deploy. I think it's subject for another issue :)
I don't know how other IDEs behave, but in PHPStorm it's enough to give static-type of method's param. If it's what you're talking about.

SenseException:
@eps90 I see you already noticed my first comments on some code :-)
Good work adding tests for those commands.

The malkusch/php-mock project is a nice help, but we should agree on, that the goal is to work toward a code, that doesn't need this anymore in the future. We may mock a behaviour away, that can differ between PHP Versions without knowing it. I don't know if we manage to remove it completly one day, but maybe it is possible. For now, we should use it. Filesystem access can someday be moved into an own Filesystem class, that can be tested separately with vfsStream.

tests/MageTest/Command/AbstractCommandTest.php
SenseException: There is no refactoring needed for using depends-annotation. When used, one of the tests doesn't need Reflection anymore but depends on the positive test result of the other one. I think the less we are using Reflection functionality together with the tests, the more stable it gets, because we don't "test" Reflection with it.

eps90:
I totally agree with wyou, @SenseException . But as I said before, I didn't want to change class' contents till we have a test complet. I think if all commands will be done (tested), we can think about refactor. In `ListCommand` there's even a fatal exception if you try to run this command without initalizing the Magallanes. I could write the tests for it, but it's topic for another PR. That's why I used `php-mock` to mock built-in functions. In the future it's obvious to have `Filesystem` class :) Little steps and everything will be shining! :)

tests/MageTest/Command/AbstractCommandTest.php
eps90: Umm ok. Sorry for my yesterday note, I was thinking of dependency injection instead of depend-annotation (I was exhausted, don't ask why).
I see advantages of such solution, but some testing philosophies say that tests should be completely independent. In the other hand these are just setters, the methods with path complexity equal 1. So well.. I need to consider it. But thanks for spotting! ;)

eps90:
And sorry again, I won't handle to make fixes today.

SenseException:
@eps90 We do this in out free time so there's no reason to apologize.

eps90:
@SenseException I fixed most of things you note. I've left only that `@depends` annotation till we find out is it really a good practice (it's not in my opinion, but it has own advantages). Instead, I moved testing getters and setters to `BaseTest` class.
Thanks for reviewing!

eps90:
@SenseException ... and I think this could be merged, because you're the only person here which I discuss about unit tests here ;)

eps90:
@SenseException fixed your notes. You can re-review it :)

SenseException:
edpauto/Magallanes@cf0f15f4ce59fbcffea02d07ef41683f5ee1495c is handled with edpauto/Magallanes@832435e07ceef70d4aa7d38f7e3e5cfeb33863d8 
edpauto/Magallanes@f8804182f5f07f86babed2b902795310190cfcfa is handled in edpauto/Magallanes@3f02ebcfe4a2878d46103d7e19ee11a2b756675e

It's ok now.

eps90:
Great, thanks a lot!

Issue #206, opened 2015-03-11
Undefined variable: currentCopy in RollbackTask.php


ParkMasakazu:
Hi, I am thankful to always use the magallanes.
Thank you for the useful tool.
Recently, I was updated magallanes from v1.0.1 to v1.0.5.
But the following error After rollback occurred in v1.0.5.
```
$ mage releases rollback --release=-1 to:production
PHP Notice: Undefined variable: currentCopy in ./vendor/andres-montanez/magallanes/Mage/Task/BuiltIn/Releases/RollbackTask.php on line 140
```
This is the happy and help solve the problem.
Please continue the development of the future also a good tool.
Thanks you.

ZesInternational:
in Mage\Task\BuiltIn\Releases\RollbackTask at line 91 add:
```php
$currentCopy = $releasesDirectory . '/' . $releaseId;
```
then it will work again.

eps90:
@ZesInternational Could you make a PR with that?

eps90:
@ParkMasakazu Thanks to @ZesInternational the issue is fixed. Please install Magallanes from `develop` branch (in composer.json you need to pass `dev-develop` as version) and check if everything works fine :)

ParkMasakazu:
Hi, It's very quickly.
I have obtained from dev-develop, worked confirmation even here.
Thanks you.

eps90:
@ParkMasakazu So, can we close the issue then?

ParkMasakazu:
OK, No problem. Thanks!

eps90:
Fixed in #209

PR #207, opened 2015-03-12
Update RollbackTask.php

ZesInternational:
Fix issue #206

### DIFF
--- a/Mage/Task/BuiltIn/Releases/RollbackTask.php
+++ b/Mage/Task/BuiltIn/Releases/RollbackTask.php
@@ -87,6 +87,8 @@ class RollbackTask extends AbstractTask implements IsReleaseAware
                         $releaseIsAvailable = true;
                     }
                 }
+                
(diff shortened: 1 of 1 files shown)


Mage/Task/BuiltIn/Releases/RollbackTask.php
```diff
@@ -87,6 +87,8 @@ public function run()
                         $releaseIsAvailable = true;
                     }
                 }
+                
+                $currentCopy = $releasesDirectory . '/' . $releaseId;
```
eps90: I think it'd be more secure if `$releases` would be `rtim`-med before appending `$releaseId`:
```php
$currentCopy = rtrim($releasesDirectory, '/') . '/' . $releaseId;
```

eps90:
Ok just one comment added.
This PR is going to be merged into `master` branch instead of `develop`. You are aware of that? :) I mean, you really wanted it to be on `master`?

PR #209, opened 2015-03-13
added $currentCopy to the RollbackTask.php

ZesInternational:
A revision of the fix for issue #206


ZesInternational:
@eps90 here is the PR in develop and with the rtrim included

eps90:
Great, can we close the other one then? This commit would be in develop stage and will be merged into master with the nearest release.

ZesInternational:
Yes, we can close the issue

eps90:
Ok so since this PR is "more correct" than the other one, I'll reopen this one and close the other.

PR #211, opened 2015-03-16
The timezone should be system dependent

MikeTralala:
If the timezone is set in the php cli ini file, mage should use that one instead of the default UTC


eps90:
Good point ;)
Please correct code style and I think we could merge it.

eps90:
Ok, merging! Thanks a lot!

PR #223, opened 2015-04-30
resultFetch is always false when owner is specified in config

isyndicate:
If the "owner" is specified in environment configuration, the resultFetch value is never changed from false to true. Therefore checking for that before setting permissions is pointless.


eps90:
Your change is ok, but I look at whole file and above the line you edited there's another `if` with exactly the same statement. Could you merge it somehow?

isyndicate:
Actually now that I look at that, we don't even need a condition on line 77. $userGroup either comes from config, or we set that in the block above (starting from line 55). Let me add to this branch and you can take a look.

eps90:
looks good for me :)

Issue #111, opened 2014-08-20
Sudo management


guepe:
in runCommandRemote  integrate the option to use sudo commande instead of using apache or root.
it could be a simple option to set and then add the sudo keyword in the command line.

Could also list all commands needed for a sudo and suggest a visudo config.

Thanks :)

drAlberT:
@eps90 if sudo is smell having to be *root* is ? ..  ;) :+1: for avoiding chown/chmod as possible, giving a set of choices to the user via config parameters: use_sudo (t/f), run_chown (t/f), and so on ..

drAlberT:
I'll be more clear .. there are some "maybe-fisrt-deploy-only" tasks (i.e. setfacl) that can only be run by root. sudo is the less smelling solution IMHO

eps90:
Why do you think that tasks like `setfacl`, `chown` or `chmod` can be run only on first time deploy?

drAlberT:
@eps90 because of the man page of `setfacl`  ;)

    The file owner and processes capable of CAP_FOWNER are granted the right to modify ACLs of a file. 
    This is analogous to the permissions required for accessing the file mode.
    (On current Linux systems, root is the only user with the CAP_FOWNER capability.)

andres-montanez:
Hi all, Version 1 of Magallanes has long reached it's end of life cycle, this feature will be available in the next mayor version.
Many thanks to all!

PR #226, opened 2015-05-15
New use-sudo option

csuarez:
Hello,

In some cases a server doesn't have a `root` user, and in order to deploy with Magallanes, an user with _sudo_ permissions is needed (e.g. where I work because a security policy servers shouldn't have a `root` user). Maybe it is not an ideal scenario, but sometimes as a developer you have to manage with this kind of restrictions. So I have added a new option (`use-sudo`) to execute all deployment commands as _sudo_. The activation of this option is straightforward, and by default is set to `false`:

``` yml
deployment:
  user: foo
  use-sudo: true
```

I've changed the _TarGzTask_ too. Because you can't do an `scp` to a folder wich needs _sudo_ permissions, when `use-sudo` is enabled, the tar.gz is previously uploaded to `/tmp` and after that it is moved to the right destination.

I hope you can find it useful :)

Regards

**NOTE**: Related issues: #111

### DIFF
--- a/Mage/Task/AbstractTask.php
+++ b/Mage/Task/AbstractTask.php
@@ -180,10 +180,6 @@ abstract class AbstractTask
      */
     final protected function runCommandRemote($command, &$output = null, $cdToDirectoryFirst = true)
     {
-        if ($this->getConfig()->deployment('use-sudo', false) === true) {
-            $command = 'sudo (' . $command . ')';
-        }
-
         if ($this->getConfig()->release('enabled', false) === true) {
             if ($this instanceof IsReleaseAware) {
                 $releasesDirectory = '';
@@ -199,7 +195,7 @@ abstract class AbstractTask
 
         // if general.yml includes "ssy_needs_tty: true", then add "-t" to the ssh command
         $ttyIsNeeded = $this->getConfig()->general('ssh_needs_tty', false)
-                        || $this->getConfig()->deployment('use-sudo', false);
+                        || $this->getConfig()->deployment('use-sudo', false)
         $needs_tty = ($ttyIsNeeded ? '-t' : '');
 
         $localCommand = 'ssh ' . $this->getConfig()->getHostIdentityFileOption() . $needs_tty . ' -p ' . $this->getConfig()->getHostPort() . ' '
@@ -212,7 +208,12 @@ abstract class AbstractTask
         if ($cdToDirectoryFirst) {
             $remoteCommand = 'cd ' . rtrim($this->getConfig()->deployment('to'), '/') . $releasesDirectory . ' && ' . $remoteCommand;
         }
-        $localCommand .= ' ' . '"sh -c \"' . $remoteCommand . '\""';
+
+        if ($this->getConfig()->deployment('use-sudo', false) === true) {
+            $localCommand .= ' ' . '"sudo sh -c \"' . $remoteCommand . '\""';
+        } else {
+            $localCommand .= ' ' . '"sh -c \"' . $remoteCommand . '\""';
+        }
 
         Console::log('Run remote command ' . $remoteCommand);
 
@@ -245,6 +246,7 @@ abstract class AbstractTask
     {
         if ($this->getConfig()->release('enabled', false) === true) {
             $releasesDirectory = $this->getConfig()->release('directory', 'releases');
+
             $deployToDirectory = $releasesDirectory . '/' . $this->getConfig()->getReleaseId();
             return 'cd ' . $deployToDirectory . ' && ' . $command;
         }
--- a/Mage/Task/AbstractTask.php
+++ b/Mage/Task/AbstractTask.php
@@ -245,7 +245,7 @@ abstract class AbstractTask
         if ($this->getConfig()->release('enabled', false) === true) {
             $releasesDirectory = $this->getConfig()->release('directory', 'releases');
             $absoluteReleasesDirectory = rtrim($this->getConfig()->deployment('to'), '/') . $releasesDirectory;
-            $deployToDirectory = realpath($absoluteReleasesDirectory . '/' . $this->getConfig()->getReleaseId());
+            $deployToDirectory = $absoluteReleasesDirectory . '/' . $this->getConfig()->getReleaseId();
             return 'cd ' . $deployToDirectory . ' && ' . $command;
         }
 
--- a/Mage/Task/AbstractTask.php
+++ b/Mage/Task/AbstractTask.php
@@ -197,7 +197,9 @@ abstract class AbstractTask
         }
 
         // if general.yml includes "ssy_needs_tty: true", then add "-t" to the ssh command
-        $needs_tty = ($this->getConfig()->general('ssh_needs_tty', false) ? '-t' : '');
+        $ttyIsNeeded = $this->getConfig()->general('ssh_needs_tty', false)
+                        || $this->getConfig()->deployment('use-sudo', false)
+        $needs_tty = ($ttyIsNeeded ? '-t' : '');
 
         $localCommand = 'ssh ' . $this->getConfig()->getHostIdentityFileOption() . $needs_tty . ' -p ' . $this->getConfig()->getHostPort() . ' '
             . '-q -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no '
--- a/Mage/Task/AbstractTask.php
+++ b/Mage/Task/AbstractTask.php
@@ -195,7 +195,7 @@ abstract class AbstractTask
 
         // if general.yml includes "ssy_needs_tty: true", then add "-t" to the ssh command
         $ttyIsNeeded = $this->getConfig()->general('ssh_needs_tty', false)
-                        || $this->getConfig()->deployment('use-sudo', false)
+                        || $this->getConfig()->deployment('use-sudo', false);
         $needs_tty = ($ttyIsNeeded ? '-t' : '');
 
         $localCommand = 'ssh ' . $this->getConfig()->getHostIdentityFileOption() . $needs_tty . ' -p ' . $this->getConfig()->getHostPort() . ' '
--- a/Mage/Task/AbstractTask.php
+++ b/Mage/Task/AbstractTask.php
@@ -244,7 +244,7 @@ abstract class AbstractTask
     {
         if ($this->getConfig()->release('enabled', false) === true) {
             $releasesDirectory = $this->getConfig()->release('directory', 'releases');
-            $absoluteReleasesDirectory = rtrim($this->getConfig()->deployment('to'), '/') . $releasesDirectory;
+            $absoluteReleasesDirectory = rtrim($this->getConfig()->deployment('to'), '/') . '/' . $releasesDirectory;
             $deployToDirectory = $absoluteReleasesDirectory . '/' . $this->getConfig()->getReleaseId();
             return 'cd ' . $deployToDirectory . ' && ' . $command;
         }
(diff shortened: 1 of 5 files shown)


eps90:
Looks quite ok. Coveralls screams but because of classes that are not tested directly (yet)
But let me point few things:
1. You've commited new `composer.lock` - why? Was there need to upgrade something?
2. There are a bunch of code style issues - did you try to analyze your code with PHPCodeSniffer or MessDetector? I can point few of them if you want :)
3. You don't need to comment out the example config value in configs - they are only for demonstration reasons.
4. I would tell something more about inconsistency in config keys naming but I can see there are lot of other config that use dash instead of underscore in config names.

csuarez:
@eps90 I will check everything you have commented ASAP :)

eps90:
Thank you! :)

Mage/Task/AbstractTask.php
```diff
@@ -194,9 +194,12 @@ protected function getParameters()
         }
 
         // if general.yml includes "ssy_needs_tty: true", then add "-t" to the ssh command
-        $needs_tty = ($this->getConfig()->general('ssh_needs_tty', false) ? '-t' : '');
+        $ttyIsNeeded = $this->getConfig()->general('ssh_needs_tty', false)
+                        || $this->getConfig()->deployment('use-sudo', false);
```
eps90: Use the indentation in the same way as you have used below for string concatenation (with dot `.`)

eps90:
Hi @csuarez !
Almost there! :) I've added you one more comment about coding style. Almost everything is brilliant, but can you tell me why did you commit the `composer.lock`?

csuarez:
Hi @eps90 ,

I was on holiday, sorry for the late response. I thought that I reverted the changes in composer.lock (it is not necessary, as you pointed it in your first response). How can I revert that change in a proper way?

BTW, maybe somewhere in the docs will be necessary to point that the deployment user will need to make `sudo` without password ([like this](http://askubuntu.com/questions/192050/how-to-run-sudo-command-with-no-password)) in order to not to stop the Magallanes deployment execution. Where could I do that?

Regards

csuarez:
Ok, I've managed to restore the composer.lock :)

Mage/Task/BuiltIn/Deployment/Strategy/TarGzTask.php
```diff
@@ -132,7 +156,10 @@ public function run()
     protected function excludesListFile($excludesFile)
     {
         $excludesListFileRsync = '';
-        if (!empty($excludesFile) && file_exists($excludesFile) && is_file($excludesFile) && is_readable($excludesFile)) {
+        if (!empty($excludesFile)
+            && file_exists($excludesFile)
+            && is_file($excludesFile)
+            && is_readable($excludesFile)) {
```
drAlberT: is_readable() Returns TRUE if the file or directory specified by filename exists and is readable, FALSE otherwise.

I think this overcheck is not needed, furthermore is there any special reason for check that it is a file(why do you avoid links?)?

Mage/Task/BuiltIn/Deployment/Strategy/TarGzTask.php
```diff
@@ -81,28 +84,49 @@ public function run()
...
             $strategyFlags = $strategyFlags['targz']['exctract'];
         } else {
             $strategyFlags = '';
         }
 
+        $sudo = $this->getConfig()->deployment('use-sudo', false);
+
+        $uploadDirectory = $deployToDirectory;
+
+        if ($sudo === true) {
```
drAlberT: if ($sudo)  should suffice AFAIK

Mage/Task/AbstractTask.php
drAlberT: tty OR ssh agent is required .. so why enforce tty?  I could want to have an ssh agent running and avoid tty for some reason ..

My opinio is to enforce as few as possible .. always .. the user can adjust the config as per his needs and environment, without being force to have an hardcoded behaviour

Mage/Task/BuiltIn/Deployment/Strategy/TarGzTask.php
eps90: well since `is_readable` covers `file_exists` and `is_file`, I'd throw them out and leave only `is_readable` here because it also checks whether the file can be opened with read permissions.

Mage/Task/BuiltIn/Deployment/Strategy/TarGzTask.php
eps90: true ;)

Mage/Task/AbstractTask.php
```diff
@@ -194,9 +194,12 @@ protected function getParameters()
         }
 
         // if general.yml includes "ssy_needs_tty: true", then add "-t" to the ssh command
-        $needs_tty = ($this->getConfig()->general('ssh_needs_tty', false) ? '-t' : '');
+        $ttyIsNeeded = $this->getConfig()->general('ssh_needs_tty', false)
```
drAlberT: tty OR ssh agent is required .. so why enforce tty? I could want to have an ssh agent running and avoid tty for some reason ..

My opinion is to enforce as few as possible .. always .. the user can adjust the config as per his needs and environment, without being forced to follow an hardcoded behaviour

drAlberT:
@eps90 please merge it .. it is extremely important for ppl like me not deploying as root ;)

Mage/Task/BuiltIn/Deployment/Strategy/TarGzTask.php
csuarez: Hello. This is like I found it, i have only change it to solve PSR-2 warnings. AFAIK, you're right.

drAlberT: yes I know it was already there :) not your bad .. but I think it is to be cleaned up :)

csuarez:
@drAlberT Normally, in a deployment scenario you'll use sudo to operate over directories where you don't have write permissions, so you can't upload files directly using scp to the targer directory. AFAIK, the best way it's to copy them to a temporary directory and afterwards move it.

http://superuser.com/questions/138893/scp-to-remote-server-with-sudo

drAlberT:
@csuarez I got your point, but I don't think it is a good practice to enforce behaviours 
In my case, for example, I set FS acl on the project dir, making all the tree writeable by the deploying user (I made this using sudo in my SetFaclTask, see my fork .. I still have to make the sudo use optional I know)

So, I think you can enable the user to choose to use a temporary dir or not using a parametr, if yes you should not hardcode the dir path, but find a way to get the system tmp dir or let the user specify it

use tmp? -> gimme tmp path -> otherwise guess the remote system $TMP ..

Hope to have been clear :)

csuarez:
@drAlberT About the local command issue... Maybe, I've missed something (sorry in advance), but the 'use-sudo' option only it's checked in runRemoteCommand function. I have added the "sudo" to the $localCommand variable because when you execute a deployment process there are a lot of chained commands using "&&" and if you add sudo to $remoteCommand, only the first command will be executed as sudo. Maybe there is a better way to do this, by it was the only I realized :sweat_smile:

csuarez:
@drAlberT Ok, I can see your point about the temporary folder problem. I would try to code it ASAP (unfortunately it has to be the next week :disappointed: ).

drAlberT:
@csuarez ok I'm busy until next week too.. then I'll try to code this feature on the remaining tasks too .. moreover there is a lot of "magic" into the deploy task now making chown and chmod that nowadays fails for me .. I'll see if using the new use-sudo switch could be the solution

have a nice WE hten .. see you next week :)

csuarez:
@drAlberT I also had the chown/chmod issues, and with use-sudo it is solved (at least for my case). Have a nice weekend too :+1:

eps90:
@drAlberT @csuarez Hey guys!
I can see everyone got a lot of work recently - me too, the whole previous week was so exhausting for me that I have completely no time to visit GitHub and, especially, Magallanes.
I've just read your conversation and concluding: (correct me if I'm wrong)
* local user should not run local commands as root - obviously ;)
* `sudo` should be prepended only for remote commands - as a conclusion from the first point
* @drAlberT I see you don't want deploy as root, but remember it's still an option and anyone who decide to deploy with `sudo` should be aware of risk
* Before first deploy Magallanes ensures that there are proper directories on target server(s) and sets permissions for deploy user. So actually @drAlberT is right and deploy to temporary directory is not necessary because all permissions are set correctly.

Did I miss something? What are we gonna do with this issue?
PS. Thanks @drAlberT and @csuarez - you make a great job in Magallanes deployment!

drAlberT:
@eps90 Of course deploying as root has to remain an option .. but please tell me, what should be the "risk of sudo"? sudo is a risk as a counterpart of unprivileged operations, but I can't see how sudo can be more dangerous of being root ...

eps90:
@drAlberT Of course by "root" I mean user with root privileges, so I treated it as an alias for `sudo` :)
In my comment I was trying to say that if someone is sure he wants to use sudo, we can't block that feature. So concluding - we can't guarantee that someone else's system won't crash after using sudo, he takes all responsibility for using that option.
E.g. capistrano has the option `use_sudo` for a long time and afaik no one complains about it :)

drAlberT:
@eps90 ah ok .. I misunderstood :)

But pay attention, I do not *want* to use sudo .. I'll be very very *if* magallanes would not try to do chown/chmod behind the curtains making my deploy to fail ...
I say that *if* one need some privilege escalation *then* we should support sudo .. but sudo should be avoided (and nowadays I can't ;( for the chown/chmod reason)

hope to have been explained :))

eps90:
@drAlberT Sorry for misunderstaning too ;)
Well I know it's hard to avoid sudo, but to keep project consistency sometimes you need to provide proper permissions to your application file system, to be usable by e.g. Apache user. Of course you can deploy it as a e.g. `www-data` but... no! :) `www-data` is an Apache user and also Apache is permitted to use it for his own reasons. I think there must be a `deployer` user with ability to use `chown/chmod` on some of your directories. This is my opinion though ;)

drAlberT:
@eps90 On my server the "webmaster" user is allowed to do anything on docroots via setfacl (and I use it on mege too .. I'm working on my fork to deploy rewrite .. take a look if you have some time ..

Hope to have a decent release in a few days :) (I'm also refactoring some existing code)

eps90:
@drAlberT Great but do not make too big pull request because they are hard to review and then hard to fix by the author ;) Big plus for you for any contribution you make! :)

mmgajic:
Hi,

What will happen with this pull request? I'm waiting for while for this feature.

Thanks for your contribution, great job!!!

drAlberT:
It seem this project is abandoned ;(

csuarez:
Hi @mihhac 

I'm (sadly) agree with @drAlberT . You can use my fork, I've been using it for months without problems.

mihhac:
That's a pity. Thanks, I think that I will try to find another solution available via composer.

misterboe:
i merged this commit on https://github.com/misterboe/Magallanes

eps90:
@csuarez Does the code on your fork work? Is it ready to merge?

drAlberT:
@eps90 is this project still alive? ..

csuarez:
@eps90 I think so (I used it on production), but some time have passed since I coded (and used it). BTW, it seems that @misterboe has been use this PR recently.

eps90:
@drAlberT I'm very sorry but I'm very very busy recently. I have a lot of work to do and I feel so guilty and ashamed I have to time to support open source projects, particularly this one. 
2 days ago I felt sick so I decided to see what changed in Magallanes since I've been off. And I found **this**, the PR that hangs unmerged since I remember.

@csuarez @misterboe Ok so let's wait for some confirmation :)

andres-montanez:
Many thanks to all!

PR #232, opened 2015-05-28
security fix, use escapeshellcmd()

drAlberT:


eps90:
Great! Thanks a lot! :)

drAlberT:
@eps90 I can confirm that applying this PR without updating commands to properly make escapes **breaks** the commands execution .. 

So you should revert or test that everything works as expected ..

eps90:
@drAlberT Could you create a new issue with link to this issue? I'll test it in my free time and such issue will be better to find.

drAlberT:
@eps90 Ok .. it could suffice to simply escape the command, without the output redirection .. but only id for "command" we intend the program name .. otherwise the command line has to be difided into command name + arguments (being escaped by escape_shell_args ) ..

drAlberT:
New issue opened as requested #237

PR #234, opened 2015-06-02
Set branch revert as the last task on post deployment

aglejberman:
Changing order of the 'scm/change-branch' task to be the last task executed on post deployment stage.
This change allows to execute current branch tasks on post deployment stage.


andres-montanez:
Many thanks!

PR #238, opened 2015-06-09
Optional timezone configuration

DaniSancas:
Reads an optional *timezone param* from `general.yml` to log with the correct datetime. Now is always writing the default `UTC` timezone, so my logs are 2h "delayed". The configuration is easy as hell:

```yaml
# /config/general.yml
# global settings
name: MyApp
email: mi@email.com
# ...more params
timezone: 'Europe/Madrid'
```

If none specified, default `UTC` will apply. The same result will apply if a wrong timezone is specified (p.e. 'asdfasd/asdfasd')

Signed-off-by: Dani Sancas <lord.sancas@gmail.com>

### DIFF
--- a/Mage/Console.php
+++ b/Mage/Console.php
@@ -100,6 +100,9 @@ class Console
 
         // Command Option
         $commandName = $config->getArgument(0);
+        
+        // Timezone for logging
+        @date_default_timezone_set($config->general('timezone', 'UTC'));
 
         // Logging
         $showGreetings = true;


DaniSancas:
Well, I've just realized that another guy did a quite similar PR: https://github.com/andres-montanez/Magallanes/pull/211

I'm closing this PR, it doesn't make sense any longer. Sorry guys, my fault ;)

eps90:
Anyway, thanks for your effort :)

PR #240, opened 2015-06-12
Improved deploy command code style

ramunasd:
Some style improvements, so far, for deploy command


andres-montanez:
Hi, Version 1 of Magallanes has long reached it's end of life cycle, so no further changes will be integreted to it, focus is for the next next mayor version.
Many thanks!

PR #242, opened 2015-07-01
Fixed "no such file" when logging is off

drAlberT:


drAlberT:
@eps90 any issue with this PR? ;)

PR #243, opened 2015-07-02
Fix PSR2

drAlberT:
Fix PSR2 compliance project wide

(run `php-cs-fix --level=psr2`)


eps90:
Thanks, @drAlberT ! ;)

drAlberT:
:+1:

PR #253, opened 2015-08-07
fixing php54 array syntax

jhuet:
That's it :)


eps90:
Hey, sorry for my absence, I was unavailable due to my private reasons. I hope I'll pay more attention to Magallanes now :) Only if I'll find some time.
Of course, merging! :)
PS. @andres-montanez Shouldn't we drop support for PHP 5.3? It's very old and not supported anymore..

jhuet:
Hey @eps90, no problem for the delay, we can't be 100% all the time :)

PR #254, opened 2015-08-07
A task to create a simple symbolic link

jhuet:
An alternative to `LinkSharedFilesTask` which allows for the creation of symbolic links for single files or directories.


eps90:
Looks great, thanks ! :)

PR #260, opened 2015-10-21
git local commands should be executed in the deployment:from directory

jhuet:
For some reasons, i [already did this change](https://github.com/andres-montanez/Magallanes/commit/a58dcbe7668dd8ebe7a5b39391d1203cfd96388b#diff-9c1fd384ca344151acde30487d87057a) for other commands in this file but not all of them...


andres-montanez:
Thanks!

PR #263, opened 2015-10-30
[FEATURE] Added "default" option to ApplyAcl task

kuborgh-mspindelhirn:


DaniSancas:
+1
Just was looking for this right now!

andres-montanez:
Many thanks!

Issue #39, opened 2014-01-08
Dry run option


seydu:
Hi. 
This seems to be the exact library I was looking for ! Thank you for that.
I used to run commands like those in Magallanes in a very awkward code, written in Shell scripts. I couldn't even get my co-workers to help me maintain the code.
The install of Magallanses using Composer is smooth, so is the configuration. 
But I have two questions:
1/ Is there a way to do a dry-run ? If yes how ?
2/ For the connection without a passphrase, how can I set the path to the private key file ? The only way I found is by defining an alias in ~/.ssh/config and using the Host alias in the environment configuration.

andres-montanez:
Hi, Version 1 of Magallanes has long reached it's end of life cycle, so no further changes will be integrated to it, focus is for the next next mayor version.
Many thanks!

PR #265, opened 2015-11-04
Add "Dryrun" mode for Rsync strategy

X-PRIME:
# Dry-run mode with Rsync strategy

Sometimes it's useful to list the files to deploy without deploy you application. That's why "dry-run" mode is here.
### How to use it ?

You just to had a "dry-run" parameter to your deploy command like this : 

```
mage deploy to:production --dry-run
```
### How about "pre" and "post" deploy tasks ?

When using dry-run, all defined tasks are skipped : 
- Pre Deploy
- On Deploy
- Post Release
- Post Deploy

But you now have a new option called "pre-dryrun" in your environment conf file where you can add all tasks before dry-run is launched.

```
tasks:
  pre-deploy:
  pre-dryrun:
  on-deploy:
  post-release:
  post-deploy:
```

Related issue : 
https://github.com/andres-montanez/Magallanes/issues/39

### DIFF
--- a/Mage/Command/BuiltIn/DeployCommand.php
+++ b/Mage/Command/BuiltIn/DeployCommand.php
@@ -130,6 +130,24 @@ class DeployCommand extends AbstractCommand implements RequiresEnvironment
     }
 
     /**
+     * Launch a Dry run command for Rsync strategy
+     * Without others sub tasks.
+     */
+    public function dryRun()
+    {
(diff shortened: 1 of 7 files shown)


wadjeroudi:
@eps90 @andres-montanez :+1:

andres-montanez:
Many thanks!

PR #268, opened 2015-11-21
Check if "chown" is available"

georgringer:
on some hosters, the command "chown" is not supported. Therefore it must be checked if it is there, otherwise no deployment is possible with magallanes


misterboe:
would that be merged? i have the same problem

qzminski:
Can this be finally merged or is the project no longer maintained? I can't deploy to some servers without that patch.

andres-montanez:
Many thanks!

PR #269, opened 2015-11-27
[FIX] Fix linked folder creation

1ma:
The LinkSharedFilesTask creates symbolic links in two steps:
1. Creates the directory that shall host the actual file/directory under shared/
2. Creates the symbolic link, using the newly created directory as its target

Step 1 works fine in the case of files because the command uses `dirname($target)` as the folder name, but in the case of directories this call actually chops the desired folder name, and the subsequent command produces a broken symbolic link.

This pull request fixes the issue by applying `dirname()` on `$target` only when appropriate (that is, when we're dealing with files, not directories).


andres-montanez:
Thanks!

PR #270, opened 2015-11-27
Use environment variables for changeing deployment and releases settings

garfieldius:
This makes the deployment configurations more flexible and easier to
reuse.

Contains unit tests for the feature.

Documentation PR is at https://github.com/andres-montanez/MagallanesSite/pull/10


andres-montanez:
Many thanks!

PR #272, opened 2015-12-03
Make the task resolution more flexible

garfieldius:
This allows to use namespace + classname combinations that integrate more
beautifully into custom packages.

Documented in andres-montanez/MagallanesSite#11


andres-montanez:
Hi, Version 1 of Magallanes has long reached it's end of life cycle, this feature will be available in the next mayor version.
Many thanks!

PR #278, opened 2016-01-06
[FEATURE] A task to copying files and folders

nolotz:
A task to copying files and folders


andres-montanez:
Still I will create this as a Feature Request for the next version, it's a very useful task!
Many thanks!

PR #279, opened 2016-01-13
Escape targz excludes to allow wildcard

marmotz:
Allow wildcard in excludes configuration:

``` yml
deployment:
  ...
  excludes:
    - ./src/*/Tests
    - docker-compose.yml
```

will generate :
`tar cfz /tmp/mageTeVWVI.tar.gz  --exclude="./src/*/Tests" --exclude=docker-composer.yml -C ./ .`


andres-montanez:
Many thanks to all!

Issue #245, opened 2015-07-06
error deploy on freebsd 


vyachin:
2015-07-06 13:34:11 -- ---------------------------------
2015-07-06 13:34:11 -- ---- Executing: $ ssh  -p 22 -q -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no user@example.ru "sh -c \"cd /home/user/www/dsf.example.ru && ln -sfn releases/20150706133402 current.tmp && chown -h user:user current.tmp && mv -fT current.tmp current\""
2015-07-06 13:34:11 -- mv: illegal option -- T
usage: mv [-f | -i | -n] [-v] source target
       mv [-f | -i | -n] [-v] source ... directory
2015-07-06 13:34:11 -- ---------------------------------
2015-07-06 13:34:11 -- FAIL
2015-07-06 13:34:11 -- Finished the Releasing

vyachin:
> uname -a
FreeBSD example.ru 8.3-RELEASE-p16 FreeBSD 8.3-RELEASE-p16 #0: Tue Apr 29 23:49:15 UTC 2014     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64

MM-Psiiirus:
Same error Solaris / SmartOS...i think T isn't even a proper parameter on any unix system , isnt it?

vyachin:
The parameter T exist only in linux version mv

MasterB:
I confirm this bug on FreeBSD

bpoiss:
+1

Najki:
I can confirm that deployment with @fetzi's [cyberhouse/Magallanes](https://github.com/cyberhouse/Magallanes) fork works on FreeBSD. Unfortunately rollback doesn't. I hope that @fetzi will fix that soon but for now, his fork seems to fix the FreeBSD issue.

andres-montanez:
Hi all, Version 1 of Magallanes has long reached it's end of life cycle, so no further changes will be integrated to it, focus is for the next next mayor version. New version doesn't have mv command so it should work with FreeBSD.
Many thanks to all!

PR #283, opened 2016-02-23
Issue #245 remove illegal option from mv command

fetzi:
On FreeBSD the `-T` option for the `mv` command is not available. Therefore removed.


fetzi:
Improved solution deletes existing symlink before creation

PR #284, opened 2016-03-01
Issue #245 remove illegal option from mv command

fetzi:


andres-montanez:
Many thanks!

PR #285, opened 2016-03-07
Change the point the current branch is saved.

MikeTralala:
This is necessary to save to reverse back to the right branch if another branch is used than 'master' and you´re currently working in the deployed branch.


andres-montanez:
Thanks!

Issue #246, opened 2015-07-06
Custom task was not found


eaglefsd:
Heyho,

I'm trying to fix my permissions with a custom task.

My code looks like this:

``` php
namespace Task;

use Mage\Task\AbstractTask;

class Permissions extends AbstractTask
{
    public function getName()
    {
        return 'Setze Berechtigungen auf 775';
    }

    public function run()
    {
        $command = 'chmod -R 775 /path/to/directory';
        $result = $this->runCommandRemote($command);

        return $result;
    }
}
```

The file is called Permissions.php and it is stored in .mage/tasks.
I call the task with `- permissions`.

Magallanes always tells me that "Task Permission was not found". What am I doing wrong?

benr77:
I have the same issue. Custom task set up and working correctly on 1.0.3

Upgraded to 1.0.4, 1.0.5 or 1.0.6 and suddenly the custom tasks cannot be found. Roll back to 1.0.3 and they are executed correctly again.

pfastre:
Same issue here. 1.0.3 works, 1.0.6 not.

tilt86:
Idem for me with 1.0.6. Not tested with version < 1.0.6

501c3:
I also find the exact same issue with 1.0.6.  Some notes:
I installed magellanes in /opt/magellanes according to the instructions for a system wide installation in the first couple of paragraphs of the website....

marcocarnazzo:
Same issue with 1.0.6.
As a workaround I converted MyCustomTask into a builtin task.

tilt86:
I used this repo https://github.com/magephp/magallanes and it works.

Amzani:
I found a workaround @eaglefsd 

Use this :
```{r, engine='bash'}
tasks:
     on-deploy:
        - scm/force-update
        - general/manually:
           - chmod -R 775 /var/www/vhosts/hassels.eu/schule.hassels.eu/
```

Amzani:
@tilt86 what's the # btw  https://github.com/magephp/magallanes  and this repo ?

1ma:
@andres-montanez @eps90 & to whom it may concern, I've submitted a PR to fix this issue. I'd be nice to have a hotfix release.

andres-montanez:
Hi all the PR has been merged, also Version 1 of Magallanes has long reached it's end of life cycle, and this feature will be available in the next mayor version and will be even better, taking full advantage of Composer.
Many thanks to all!

PR #286, opened 2016-03-18
Fix regression introduced in 1.0.4: custom tasks not found

1ma:
Addresses issue #246 

Commit [892a60c](https://github.com/andres-montanez/Magallanes/commit/892a60c57461176dd7acc60d4f10c1291788aba2) introduced a change in the `mage` script by which the `\Mage\Autoload` class would never be registered in the global namespace when the `{project_root}./vendor/autoload.php` already existed (which usually does).

Since this class MUST be registered in order to find any tasks defined inside `{project_root}/.mage/tasks` I've moved this snippet outside the `else` clause. Let there be custom tasks again.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,10 +1,11 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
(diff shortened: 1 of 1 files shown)


leofeyer:
You are most likely using an optimized Composer autoloader, which has been built before you have added your commands. Try running `composer dump-autoload`.

robertosanval:
I get the not found error when I try to create my custom task. I've tried with 1.03 to 1.0.6 versions, all with the same result :(

andres-montanez:
Thanks!
Creating custom tasks will be MUCH easier in the next mayor version.
Stay tuned!!

PR #290, opened 2016-04-11
Allow `deployment -> source -> temporal` to be absent in the config

Kevinrob:
Without this, we have to set an empty value like this for using default `tmp` on the system:

```
deployment:
  source:
    temporal: ~
```


andres-montanez:
Thanks!

Issue #292, opened 2016-04-15
Notification email reports SUCCESS even if deployment failed


natechicago:
Currently, if deployment tasks fail but post-deployment tasks succeed, the CLI tool reports the deployment as a failure, but the notification email reports the deployment as a success.  This is because `DeployCommand::runNonDeploymentTasks()` resets `self::$failedTasks` to zero.  That static variable is then used further downstream to determine whether or not the email notification should consider the deployment a success.  I believe the `self::$deployStatus` variable should be used for this purpose instead.

andres-montanez:
Hi, Version 1 of Magallanes has long reached it's end of life cycle, so no further changes will be integrated to it, focus is for the next next mayor version.
Many thanks!

boroth:
I'm seeing this in my v5 deployments as well, any tips on why that might be occurring? For example, a symfony/doctrine migration was formatted incorrectly and throws a 500 error, but Mage is still reporting it as a success? Also seen instances where remote commands/scripts fail with exit code 0 and Mage still thinks they're successful 🤷🏾

Running a symfony-cmd as follows:
```
- symfony-cmd: { name: 'symfony/cmd: phoenix:migrations:migrate', cmd: 'phoenix:migrations:migrate --no-interaction' }
```

PR #293, opened 2016-04-15
292: Notification email reports SUCCESS even if deployment failed

natechicago:
Fixing the bug reported in #292


andres-montanez:
Thanks!

PR #297, opened 2016-07-14
Enable to configure strict host key checking [closes #105]

### LINKED ISSUE #105: Allow users to enable host key checking
While the default makes sense, this should be configurable.

voda:
Software should be secure by default, so the default option is to enable
strict host key checking. When lowering security, the user should be
aware of it and it should be his choice.

### DIFF
--- a/Mage/Config.php
+++ b/Mage/Config.php
@@ -389,6 +389,16 @@ class Config
     }
 
     /**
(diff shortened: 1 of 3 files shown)


andres-montanez:
Hi, Version 1 of Magallanes has long reached it's end of life cycle, this feature will be available in the next mayor version.
Many thanks!

PR #298, opened 2016-07-21
[FIX] Removed second argument for Console::log() method.

zomberg:
It was removed from this method earlier.


andres-montanez:
Thanks!

PR #299, opened 2016-07-21
[FIX] Removed unnecessary argument from calling Parser::isStringUnIndentedCollectionItem() method

zomberg:
Looks like it was refactored before.


esmiz:
Hi friend! 

I've found some information about a  very beutifulplace,  just  take a look here  <http://fundexesly.rabidorangutan.com/e6bzzrx>

All best, esmiz

andres-montanez:
Many thanks!

PR #302, opened 2016-07-21
[FIX] Moved cleanUpReleases method to parent class.

zomberg:
It uses in GitRemoteCacheTask.


andres-montanez:
Many thanks!

PR #311, opened 2016-11-23
[Feature] symfony2/apply-facls task

delbio:
This is a Specific Symfony2/3 setfacl task.
Follow [Symfony file_permission guide](http://symfony.com/doc/current/setup/file_permissions.html#using-acl-on-a-system-that-supports-setfacl-linux-bsd)


andres-montanez:
Many thanks!

PR #326, opened 2017-01-30
Reading a default ssh user while executing remote command

piomet1:


andres-montanez:
Hi, thanks for the fix!
Regards.

Issue #329, opened 2017-01-31
Composer path configuration per Environemt


ufeldmann:
Wouldn't it be better to set the individual composer path per environment instead of setting it globally?

andres-montanez:
Hi!
Ideally composer would be installed globally, but I understand that's not always the case, so yes, it makes sense to allow a per environment option too.
Thanks! I will try to squees this on the next release.

Toflar:
Same here, this is a big problem. The environments might also be split over different hosts or physical servers so this is really need. I [quickly created a PR](https://github.com/andres-montanez/Magallanes/pull/330) for this, hoping that new versions will be released more often than in the past ;-)

andres-montanez:
Hi @Toflar , thanks for the PR, I will take a look at it!
And yes, the idea is to release more often :D

Toflar:
Merged for v3.1, can be closed ;)

andres-montanez:
3.1 will be released end of the month, luckily this is the shortest month! ;)

PR #330, opened 2017-02-01
Added support for composer environments

Toflar:
I figured I need something that's used in other tasks too so I also built a small helper method in the `Runtime` class :-)

PR for https://github.com/andres-montanez/Magallanes/issues/329

### DIFF
--- a/src/Runtime/Runtime.php
+++ b/src/Runtime/Runtime.php
@@ -230,6 +230,27 @@ class Runtime
     }
 
     /**
+     * Shortcut to get the the configuration option for a specific environment and merge it with
+     * the global one (environment specific overrides the global one if present).
+     *
+     * @param      $key
+     * @param null $defaultEnv
+     * @param null $defaultConfig
+     *
+     * @return array
+     */
+    public function getMergedEnvAndConfigOption($key, $defaultEnv = null, $defaultConfig = null)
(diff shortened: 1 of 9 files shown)


src/Runtime/Runtime.php
```diff
@@ -230,6 +230,27 @@ public function getEnvOption($key, $default = null)
...
+     * Shortcut to get the the configuration option for a specific environment and merge it with
+     * the global one (environment specific overrides the global one if present).
+     *
+     * @param      $key
+     * @param null $defaultEnv
+     * @param null $defaultConfig
+     *
+     * @return array
+     */
+    public function getMergedEnvAndConfigOption($key, $defaultEnv = null, $defaultConfig = null)
```
andres-montanez: @Toflar please, rename the method to `getMergedOption` so it's not too bulky, and only use one $default.
Thanks!

andres-montanez:
Hi @Toflar the PR is fantastic, just two things.
The name of the new method if too bulky, and then please do the PR to the nostromo branch.
Everything else is great.

Thanks!

Toflar:
Updated. Also I figured that `null` as default never makes sense for that method so I defaulted it to `[]`. Want me to squash the commits or will you do on merge? :-)

PR #333, opened 2017-02-02
Update example-config.yml

MM-Psiiirus:
Because this is a production environment example I would recommend to use the symfony prod env as well.


andres-montanez:
Hi @MM-Psiiirus , well spotted!
Could you please make the PR to the "nostromo" branch?
Thanks!

Issue #340, opened 2017-02-04
Allow `fs/remove` to remove folders


thePanz:

mbaeuerle:
I used a kind of hack to delete directories: 
```
        - fs/remove: { file: '-r public_bak' }
```

thePanz:
Nice, I guess it could be considered a "bug" as the filename is not escaped during the command execution :)

mbaeuerle:
Let's call it a helpful bug 😋

andres-montanez:
LOL, "it's not a bug, it's a feature"

PR #341, opened 2017-02-04
Allow fs remove recursive

thePanz:
See #340

### DIFF
--- a/src/Task/BuiltIn/FS/RemoveTask.php
+++ b/src/Task/BuiltIn/FS/RemoveTask.php
@@ -28,7 +28,7 @@ class RemoveTask extends AbstractFileTask
     public function getDescription()
     {
         try {
-            return sprintf('[FS] Remove "%s"', $this->getFile('file'));
+            return sprintf('[FS] Remove %s"%s"', $this->getRecursive(), $this->getFile('file'));
         } catch (Exception $exception) {
             return '[FS] Remove [missing parameters]';
         }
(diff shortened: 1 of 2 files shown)


andres-montanez:
Hi!
I deliberately didn't implement a recursive delete, it's too risky.
The idea of the FS tasks is to move/copy/remove configuration files needed for a deployment.

I understand that it may be needed, for example remove a cache directory (when not using releases), so perhaps the best option here is to create a new task "remove-dir" which will test that the target is a directory.

thePanz:
Why would it be dangerous? just the rsync command is even replacing files, why removing a folder is considered harmful?

mbaeuerle:
I think there is a need for parameterization in general! For example:  I need `cp -a src/. dest/` to move whole directories. 
So an optional flags parameter would be nice (defaulting to the current parameters) like: ` - fs/copy: { from: 'from/.', to: 'to/', flags: '-a' }`
A deployment script should be flexible and providing the tools for it is not harmful in my opinion.

andres-montanez:
+1 to adding flags!

@thePanz regarding the harmfulness of the command, think about a user that misconfigures the command and ends with something like `rm -rf /`, that's my main concern.

On the other hand I think @mbaeuerle nailed it with the flags, we are already using that in many other commands, and the user will have to thoughtfully add it.

I've created issue #344 so I can address this in the next release ;)

PR #342, opened 2017-02-04
Rsync task: ignore empty 'exclude' parameters

thePanz:
When empty lines are left in the "exclude" list, RSync is still sending `--eclude=` with empty excluded items


andres-montanez:
Well spotted!! Could you please also add it to the Tar task?
Thanks!

thePanz:
Updated

andres-montanez:
Thanks!

Issue #316, opened 2017-01-08
[Tasks] Pre-register Custom Tasks


andres-montanez:
Allow to pre-register custom tasks on the config file so if the task is used many times on the same project then it can be invoked by its "name" and not by it's classname

Example:
```yaml
magephp:
    environments:
        production:
            post-deploy:
                - do-magic
        staging:
            post-deploy:
                - do-magic
        testing:
            post-deploy:
                - do-magic
        demo:
            post-deploy:
                - do-magic
    custom_tasks:
        - AppBundle\Library\Deployment\DoMagicTask
```

```php
namespace AppBundle\Library\Deployment;

use Mage\Task\AbstractTask;

class DoMagicTask extends AbstractTask
{
    public function getName()
    {
        return 'do-magic';
    }

    public function getDescription()
    {
        return '[Custom] Doing some magic';
    }

    public function execute()
    {
        /* some awesome magic happening here */
    }
}
```

At init time all registered custom tasks will be instantiated and added just like the built in tasks.
This should make the config file more readable, and also if the class needs to be moved then you just only have to change it in one place.

dn5:
Looking forward to see this. Did you think about the way you are planning to implement this?

andres-montanez:
Hi @dn5 like in the issue description, and load the tasks at runtime after the built in tasks.

andres-montanez:
Implemented in version 3.2

Issue #347, opened 2017-02-07
Allow user to list built-in/custom tasks


duraki:
This could be a really great option for debugging the deployment procedure, and it's usage would be fairly simple. We would need a command for example `mage:tasks:builtin` to list all built-in tasks possibly with description and usage, and `mage:tasks:custom` to list all custom tasks.

We could further extend the task listing by allowing dummy tests to be made directly instead of writing unit tests specifically.

A lot of times while writing custom tasks, I have to test deployment on my local environment (when not using dry-run), while instead, I could run a task and magically get the output if the task passed or failed because of syntax error or similar.

andres-montanez:
Sounds cool!

andres-montanez:
We'll rely on the documnetation page for the list of built in tasks.
For custom task we'll rely on the developer keen eyes.

PR #352, opened 2017-02-08
[#316] Added the support for custom tasks

qzminski:
See https://github.com/andres-montanez/Magallanes/issues/316

### DIFF
--- a/tests/Resources/testhost-custom-task.yml
+++ b/tests/Resources/testhost-custom-task.yml
@@ -14,7 +14,9 @@ magephp:
                 - composer/install
                 - composer/dump-autoload
             on-deploy:
-                - Mage\Tests\Task\CustomTask
+                - custom
             on-release:
             post-release:
             post-deploy:
+    custom_tasks:
+        - Mage\Tests\Task\CustomTask
--- a/src/Task/TaskFactory.php
+++ b/src/Task/TaskFactory.php
@@ -42,6 +42,7 @@ class TaskFactory
     {
         $this->runtime = $runtime;
         $this->loadBuiltInTasks();
+        $this->loadCustomTasks();
     }
 
     /**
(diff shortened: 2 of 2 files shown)


qzminski:
By the way this PR I think that an exception should be thrown if the task does not extend ```AbstractTask``` or otherwise the task would not be registered and executed – and no error thrown. We should notify the developer immediately that the registered task is not executable.

Toflar:
I'd try to have some sort of interface so it's not even required to add anything at all anymore ;-)
Imagine a `Mage\Task\AutoloadInterface` with no methods at all (or we could even make an official one that includes `execute()`, `getName()` and `getDescription()` which would be a good idea anyway).

All you need to do is something like

```php
foreach (get_declared_classes() as $class) {
    if (is_a($class, AutoloadInterface::class, true)) {
        // load me
    }
}
```

andres-montanez:
Hi, yes I was thinking of adding an interface in case the abstract class is too bulky.
Also agree with about throwing an error if the custom can't be loaded.

The problem with get_declared_classes() is that it will only return the used classes (or those requested), iterate over the custom_tasks collection will be still needed and a way to get them declared, like class_exists does, which pings the autoloader.

Toflar:
Ah of course, you're right. Thinking about using Composer to provide that information but it would require to have installed the dependencies using `--optimize-autoload` to have an array of available classes. I never do that during development...

qzminski:
Then how about providing the task-list classes in the ```custom_tasks``` section? See example:

```yml
custom_tasks:
    - 'FoobarLibrary\TaskList'
```

```php
namespace FoobarLibrary;

class TaskList extends SomeTaskListInterface
{
    public function getTasks()
    {
        return [
            'FoobarLibrary\Task1',
            'FoobarLibrary\Task2',
        ];
    }
}
```

Toflar:
Imho it should be named `TaskProviderInterface` and thus the key should be `task_providers:` instead of `custom_tasks`. I think that would be cool.

andres-montanez:
It sounds interesting and full of potential, BUT that won't be easy to read. You will end up with two configuration files, and one is a class. It will be easy to loose track of which tasks are being loaded and which doesn't.

Also let's not get out of the scope, the idea is an easy way to define custom tasks and save writing the class name each time.

We can split this and also add a feature for a Task Provider which could return tasks accordingly to the environment. For example if env is Production, return ['pre-deploy' => [tasks], 'post-deploy' => [tasks]], that seams a lot more useful. You will end up with a split config but at least it could be dynamic :)

qzminski:
That's true but I think it does not much matter which task is loaded until it's actually used. Also the full list of tasks could be printed once #347 is implemented.

Indeed the idea of custom tasks we should leave as is and this new feature would go to the task providers.

As a task provider I would rather see a class that returns the list of available classes but does not decide about the stage they are executed in. This is too environment specific to be abstracted to some reusable portion of code. We've came up with this idea as we have an extra library (included via composer) for some extra deployment tasks and it would be easier to just provide ```custom/foobar``` task in the config instead of the full class name (which can change anytime contrary to the rather static task name/id).

andres-montanez:
@qzminski I think we can meet both requirements, the provider can pre register the classes, but also could inject them if configured to do so. I know that I will benefit from the later ;)

qzminski:
Oh well, sounds good :) Shall I provide a PR?

Toflar:
I think we have to differentiate here. A task provider makes a lot of sense if you have an extra library that you reuse a lot. `composer require` and adding the task provider class to `.mage.yml` would be all there is to do then. That's pretty cool. Still registering them globally makes sense if you have local, project specific tasks.

tests/Resources/testhost-custom-task.yml
```diff
@@ -14,7 +14,9 @@ magephp:
                 - composer/install
```
Toflar: Sorry but this test is nonsense. It increases coverage but it does not test anything at all. You might as well just drop it. Please add a test that checks if `CustomTask` was really registered :)

andres-montanez:
Closing the PR due to coverage drop.
Feature will be addressed in v3.2.
Regards.

Issue #332, opened 2017-02-01
Add a composer selfupdate task


Toflar:
I think it's pretty nice when you deploy and Composer updates itself.
However, it doesn't really make sense to do that for every single deployment. I think it would make sense to check if the Composer version is older than <configurable> days.

We could use `composer --version` that outputs the date for that.
WDYT?

You can assign this to me if you like the idea. I'll happily provide a PR for this feature :)

aschempp:
I already did something like that in v1 by simply adding a custom task that does composer self-update. I only run this in dev mode because I usually deploy to dev before prod.

Toflar:
Same here, but my production system does not have the same instance as my dev system. Therefore I need to update composer also on my prod release.

andres-montanez:
Hi! It's an interesting feature, and yes I don't see another way to know how old the release is.
If it's more than XX days, you can run a self-update, otherwise SKIP the task.
Also the user must be warned (in the docs) that a self-update may require special privileges.

@Toflar feel free to do a PR for this, there is no rush ;)

Toflar:
I'll work on it as soon as the other PR is merged into nostromo so I can reuse the `getMergedOption()` ;-)

andres-montanez:
PR merged ;)

Toflar:
Closing in favor of my PR :)

PR #355, opened 2017-02-10
Added composer selfupdate command

Toflar:
This is a PR for https://github.com/andres-montanez/Magallanes/issues/332.

Provides `composer selfupdate` which is not executed if the version is not older than `30` days which is configurable through the `days` option 😎


andres-montanez:
Hi! Thanks for the PR.
I've already merged it, and made a few changes.
It will be available in version 3.1, scheduled for release at the end of the month.

Regards!

Issue #336, opened 2017-02-03
Exec or equivalent through environment config


dn5:
Since we now have `filesystem` functions through configuration, I'd like to hear an input regarding the equivalent for `exec(ution)`. A lot of enterprise or large-scale projects have for example a binary applications that does some things quite necessary for a specific environment. Currently, this is at best dealt through custom task. 

I'd prefer a choice through config (`.mage.yml`) file as:

```
magephp:
    environments:
        production:
            post-deploy:
                - exec: { 'Return jobs status', './bin/supervisorctl status all > deploy_status.txt' }
                - exec: { 'Restart Supervisor jobs', './bin/supervisorctl restart all' }
```

My design would prefer a description of the execution command, followed by the second argument - the actual execution.

This of course should be defined by collective voting here, on whats the prettiest syntax to use. The following command could then be extended to various different settings and functionalities.

andres-montanez:
HI @dn5 it's an interesting idea! I will take a look at it.

dn5:
Glad to hear! I hope to get a spare time and possibly send a pull request. Looking forward for new versions, you are doing Gods work :)

dn5:
Note: This is low priority task so merge can wait for further version.

Toflar:
@dn5 enjoy: https://github.com/andres-montanez/Magallanes/pull/356 ;)

dn5:
Nice. I'd like to see your PR merged in following versions.

andres-montanez:
Hi everyone, @Toflar PR has been merged and the feature will be available in v3.1.
Thanks everyone!

PR #356, opened 2017-02-10
Added task to execute arbitrary commands

Toflar:
This is a PR for https://github.com/andres-montanez/Magallanes/issues/336.

Usage:

```yml
- exec { cmd: 'my command' }
```

If you like, you can pass a custom description like so:

```yml
- exec { cmd: 'my command', descr: 'My command does this and that' }
```

Also you can `jail` it (which forces remote execution) and adjust the `timeout`. 😎


andres-montanez:
Hi, tanks for the PR. I've tweaked it a bit.
Cheers!

PR #357, opened 2017-02-15
[FEATURE] allow exclude definition global

bennsel:
Hi,

I've allowed the exclude definition to be globally defined. My Configuration uses on every environment the same excludes and so i must implement duplicate code. With this feature i define the excludes only one time. 

Now its both possible global and environment specific

Example: tests/Resources/global-exclude.yml

### DIFF
--- a/tests/Command/BuiltIn/DeployCommandMiscTasksTest.php
+++ b/tests/Command/BuiltIn/DeployCommandMiscTasksTest.php
@@ -50,9 +50,27 @@ class DeployCommandMiscTasksTest extends TestCase
         $this->assertEquals(0, $tester->getStatusCode());
     }
 
-    public function testComposerFlags()
+    public function composerFlagsDataProvider() {
+        return [
+            'default' => [
+                'yml' => __DIR__ . '/../../Resources/composer.yml'
+            ],
+            'global exclude' => [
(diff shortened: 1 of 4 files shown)


andres-montanez:
Hi, thanks! Looks pretty handy, I will review it and schedule it for v3.2.
Cheers!

tests/Command/BuiltIn/DeployCommandMiscTasksTest.php
```diff
@@ -50,9 +50,27 @@ public function testSymfonyEnvironmentConfiguration()
         $this->assertEquals(0, $tester->getStatusCode());
     }
 
-    public function testComposerFlags()
+    public function testComposerFlagsDataProvider() {
```
thePanz: Non-test functions should not start with "test" prefix :)

bennsel: totally right. i've correct it

andres-montanez:
Hi, thanks for the PR. I've merged but then noted you have done it to the `master` branch, so I had to revert it and do it by hand to the correct branch which is `nostromo`.

So, thanks for the PR and the improvement but for further PRs please do it always to the `nostromo` branch.

Regards.

Issue #361, opened 2017-03-01
equivalent to "from:"


Havrin:
I started a new project with the new v3 of magallanes. Maybe I am missing something, I tried to copy and change my config of my old v1 projects to v3, but I am missing the `from:` parameter. I have my grunt, composer etc. stuff on root-level and a folder "http" with my project files I want to deploy. On v1 I used `from: ./http` in the configuration, now all vendor, node_modules files etc. are deployed too.

andres-montanez:
Hi! That option was removed from V3.
Your .mage file should exists in the directory from where your deployment dir starts.
By your comment above, you should have your .mage.yml inside the http directory.

Regards.

PR #362, opened 2017-03-06
[#361] Add tar_from parameter to deploy specific directories

tsdevelopment:
Equvalent 'from' parameter as in version 1 to deploy from a specific directory instead of the current root directory.

*Usage:*

```magephp:
    environments:
        development:
            tar_from: /directory/to/deploy
            [...]
```


andres-montanez:
Disregarding this PR in favor of #365 which considers Tar and Rsync.
Thanks!

PR #364, opened 2017-03-08
[FEATURE] Added support for specifying task timeout in composer install task

Dgc2002:
It's not unusual, depending on your environment, for Composer's install command to exceed the default 120 second timeout value. This allows you to configure that value and keeps the default 120 second value otherwise.   
  
Usage:  
  
```
magephp:
  environments:
    production:
      pre-deploy:
        - composer/install: { timeout: 600 }
      [...]
```


andres-montanez:
Thanks!

PR #365, opened 2017-03-14
[FEATURE] Adds from paramter for rsync and tar tasks

nolotz:
I saw that the "from" parameter got lost since the rewrite of the mage, but IMHO this is a very useful setting which shouldn't be removed.


andres-montanez:
Hi, thanks for the PR!
Next time please add a couple of tests so we can make sure the feature works as expected.

This will be available in the next release.

Regards.

PR #370, opened 2017-03-23
Feature/abstract strategy

yamilovs:
Just reformat code for working with abstract class, instead of code duplicate.


yamilovs:
Decreasing of code coverage can not depend on my commits because abstract class do the same things as it was.

andres-montanez:
Hi, thanks for the PR, but the changes on the deploy strategy will be handled further in the road map, now is too soon.

Thanks!

PR #372, opened 2017-03-27
Increase timeout for Composer task

aschempp:
The default composer:install will only run for 120 seconds which is not enough on all hosts…


Dgc2002:
Check out pull request #364

aschempp:
👍 

closed in favor of #364

PR #382, opened 2017-05-20
Add build in npm install task

GM-Alex:
Npm is a standard tool for web application development and required for grunt and bower so it should be integrated. This PR adds a build in task for npm.


b3nl:
What happens if there is no global install of the npm?

GM-Alex:
Like you can see at the code the task will return false because it wasn't successful and phpmage will show you the tasked failed. Like for the other build in tasks (git, ssh and so on).

b3nl:
In general i agree with you and i hope, that there are no admins/ops doing "sudo npm" ... but there could for example.

PR #383, opened 2017-05-20
Add build in grunt task

GM-Alex:
Grunt is a standard tool for web application development so it should be integrated. This PR adds a build in task for grunt.


b3nl:
What happens without a global install of grunt?

GM-Alex:
Like you can see at the code the task will return false because it wasn't successful and phpmage will show you the tasked failed. Like for the other build in tasks (git, ssh and so on).

b3nl:
I don't like installing everything in the global namespace, i think you too. Grunt is one of this things.

PR #385, opened 2017-05-20
Feature/ssh base64 key task

GM-Alex:
Sometimes you don't want to add your deployment ssh key to your server or if you using such systems like bitbucket pipelines you don't want to add it to your docker container. This provides a method to create from a base64 encoded ssh key a ssh key for the current user. The base64 key can be added directly to your yaml using `base64Key` see the following example:

```yaml
pre-deploy:
  - ssh/create-key-from-base64: { base64Key: 'BASE64ENCODEDSTRING' }
```

or will be taken from an environment variable using `base64KeyFromEnvVar`. Furthermore you can choose a file name for the key with `keyFileName` See the following example.

```yaml
pre-deploy:
  - ssh/create-key-from-base64: { base64KeyFromEnvVar: 'SSH_KEY_ENV_VAR', keyFileName: 'custom_key' }
```

### DIFF
--- a/src/Task/BuiltIn/Ssh/CreateKeyFromBase64Task.php
+++ b/src/Task/BuiltIn/Ssh/CreateKeyFromBase64Task.php
@@ -1,0 +1,63 @@
+<?php
+/*
+ * This file is part of the Magallanes package.
+ *
+ * (c) Andrés Montañez <andres@andresmontanez.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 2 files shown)


andres-montanez:
This is quite a hack!
Also risky, because you could overwrite a valid key.
Keys should be outside the configuration, you currently can specify a key file if needed in the flags of the ssh configuration
https://www.magephp.com/#ssh

Cheers!

PR #386, opened 2017-05-29
[FEATURE] Add timeout option for assetic:dump task

tlode:
Adds a timeout option for the assetic:dump task.

This gives the possibility to overwrite the default timeout (120 seconds) for a long running assetic:dump processes.


andres-montanez:
Thanks!

PR #390, opened 2017-06-07
Allow overriding a binary

Toflar:
Hey Andres,

I currently encounter a typical BSD tar/GNU tar issue. The `exclude` option does not work exactly the same which is why you have to specify the exclude in a way that they match your current tar settings. I came up with the idea to have a simple `binary-map`.
So now I can install e.g. `gnu-tar` (or `gtar` etc.) next to my `tar` and specify a mapper:

```yaml
magephp:
    binary-map:
        tar: gnutar
```

That allows me to have different `exclude` definitions on different machines I'm working on. Happy to hear your feedback on this.

### DIFF
--- a/src/Runtime/Runtime.php
+++ b/src/Runtime/Runtime.php
@@ -156,6 +156,26 @@ class Runtime
     }
 
     /**
+     * Retrieve a binary and check if it was changed in the "binary-map" configuration.
+     * This allows e.g. to override "tar" by "gnutar" etc.
+     *
+     * @param string $binary
+     *
+     * @return string
+     */
+    public function getBinary($binary)
+    {
+        $map = $this->getConfigOption('binary-map', []);
+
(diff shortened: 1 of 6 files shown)


aschempp:
😍 this would also allow to specify a path for binaries if they are not found in the standard locations.

andres-montanez:
Hi, thanks for the proposal, I've not accepted the PR but I've implemented the feature in another way. You will be able to define the binary path of the tar for creation and extraction. Taking your case in consideration perhaps you launch Magallanes in BSD but your destination hosts are GNU, or vise versa.

Regards!

Toflar:
Thank you :) This is just a reminder that `tar_path` should be documented on magephp.com :-)

andres-montanez:
Hi, thanks, it's already documented :)

Toflar:
Ah, I guess I searched for the wrong keys. Sorry, my bad :D

PR #392, opened 2017-07-05
[FEATURE]Support colon notation in host config definition

WorldWarIII:
Support configure different ports on same host in `host:port` notation.


andres-montanez:
Thanks!

PR #394, opened 2017-07-24
Fix getWorkingHost to return only host part

WorldWarIII:
Working host must return only `host` part.


WorldWarIII:
Can you please merge this improvement. 
Previous pull request will not work correctly without this one. Thx

WorldWarIII:
Any news on this one?

andres-montanez:
Will be solved in 3.4

PR #402, opened 2017-11-15
Added an ssh command to allow for faster ssh connections

Toflar:
Hey Andres,

I often find myself wanting to ssh to a certain environment because I quickly need to check something. When you have a lot of environments targeting different servers and using different ports and public keys that can be quite annoying because you need to copy&paste the information.
I thought there must be a better way because all of it is configured in `mage.yml` so why not reuse it? 😄 
I've invented a `mage ssh` command you can use as follows:

* `mage ssh production`, connects to the `production` environment using the configuration of that environment and the first host in the `hosts` definition.

If you want to connect to a different host, you can either specify it using a string or the array index. So given the config

```yaml
hosts:
    - webserver1
    - webserver2
    - webserver3
```

and you want to ssh to `webserver3`, you can work with both of these:

* `mage ssh --host=webserver3 production`
* `mage ssh --host=2 production`

No unit tests there because I really don't know what I should test here. TTY testing is very hard and probably not worth the pain because that command is just a helper and not a deployment relevant piece of code at all.

### DIFF
--- a/src/Command/BuiltIn/SshCommand.php
+++ b/src/Command/BuiltIn/SshCommand.php
@@ -1,0 +1,95 @@
+<?php
+/*
+ * This file is part of the Magallanes package.
+ *
+ * (c) Andrés Montañez <andres@andresmontanez.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Mage\Command\BuiltIn;
+
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Mage\Command\AbstractCommand;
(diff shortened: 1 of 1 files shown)


src/Command/BuiltIn/SshCommand.php
```diff
@@ -0,0 +1,95 @@
...
+            if (!isset($hosts[$host])) {
+                $output->writeln(sprintf('<error>%s</error>', sprintf('The host key "%s" does not exist in environment "%s".', $host, $env)));
+
+                return 1;
+            }
+
+            $host = $hosts[$host];
+        }
+
+        $cmd = sprintf('ssh -p %d %s %s@%s', $sshConfig['port'], $sshConfig['flags'], $user, $host);
```
phansys: "%u" should be more accurate for "port" placeholder.

andres-montanez:
Hi, although it looks like a nice feature, it goes beyond the scope of Magallanes.
Thanks!

PR #407, opened 2017-12-12
Make tar commands timeout configurable

humbkr:
**Use case:** Deploying a big Drupal project using releases.

**Problem:** tar commands fail because of the low value of the specified timeout in code.

**Solution:** if you set a timeout on built-in commands run by magallanes, you should also add the possibility to override this timeout, as some projects may need more time to tar / copy / untar the source code.

I added a _tar_timeout_ parameter for the commands run by the built in tar task, that can be specified alongside the other tar parameters:

```
environments:
        production:
            tar_timeout: <your value in seconds>
```

Note: the tar task is not only doing the tar / untar, but also the scp copy of the archive to the environment server, the _tar_timeout_ parameter will also impact the timeout of the scp command, by doubling the value of the parameter. For example if the parameter is set to 400 seconds, the timeout for the tar / untar commands will be 400 and the timeout for the scp command will be 800 seconds.

### DIFF
--- a/src/Task/BuiltIn/Deploy/Tar/CopyTask.php
+++ b/src/Task/BuiltIn/Deploy/Tar/CopyTask.php
@@ -45,7 +45,7 @@ class CopyTask extends AbstractTask
 
         $tarPath = $this->runtime->getEnvOption('tar_extract_path', 'tar');
         $flags = $this->runtime->getEnvOption('tar_extract', 'xfzop');
-        $timeout = $this->runtime->getEnvOption('tar_timeout', 600);
+        $timeout = $this->runtime->getEnvOption('tar_timeout', 300);
         $targetDir = sprintf('%s/releases/%s', $hostPath, $currentReleaseId);
 
         $tarLocal = $this->runtime->getVar('tar_local');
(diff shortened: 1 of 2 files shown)


andres-montanez:
In 3.4 there won't be a limit for tar and rsync.
Regards.

fritzmg:
Has this timeout been reintroduced in `4.0.0`? I just got the following error message while trying to do a deploy for the first time:
```
The process "tar cfh --force-local C:\Users\…\AppData\Local\Temp\magE104.tmp --exclude=".git" --exclude="./var/*/*" --exclude="./node_modules" --exclude="./pub/static/frontend" ./" 
exceeded the timeout of 300 seconds.

fritzmg:
As far as I can see, the default is still `300` seconds in Magallanes 3.4 too. Imho this either needs to be configurable, or it should be set to a very high number for these tasks (since you cannot _not_ set a timeout for a Symfony Process).

Hpxtech:
I know this is an old issue, but in v4.0, i still have the problem too.
I found this relatively old PR, that seems to resolve the problem with 2 lines and one environnement variable (tar_timeout), and it seems to work well :
[https://github.com/andres-montanez/Magallanes/blob/612e71e4510c84677c2bd9aa0a674d28c1b4dc18/src/Task/BuiltIn/Deploy/Tar/CopyTask.php](https://github.com/andres-montanez/Magallanes/blob/612e71e4510c84677c2bd9aa0a674d28c1b4dc18/src/Task/BuiltIn/Deploy/Tar/CopyTask.php) 
I'm not familiar with github, maybe it could be merged into the v4 ?

And great job with this magallanes, it's very useful to me !

PR #413, opened 2018-01-26
Updates to symfony 4.0

JulianStier:
Migration preparations for symfony 4.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 /vendor/
 /build
 composer.lock
+
+# IDE-specific
+/.idea/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: php
 php:
-  - '5.5'
-  - '5.6'
-  - '7.0'
   - '7.1'
 
 install:
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "symfony/event-dispatcher": "^4.0",
     "symfony/finder": "^4.0",
     "symfony/yaml": "^3.0",
-    "symfony/process": "^3.0"
+    "symfony/process": "^4.0"
   },
   "require-dev": {
     "phpunit/phpunit": "4.8.*",
--- a/composer.json
+++ b/composer.json
@@ -15,9 +15,9 @@
     "php": ">=5.5.9",
     "monolog/monolog": "^1.0",
     "symfony/console": "^4.0",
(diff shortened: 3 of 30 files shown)


.gitignore
```diff
@@ -1,3 +1,6 @@
 /vendor/
 /build
 composer.lock
+
+# IDE-specific
```
phansys: Why is this change required? Please, add environment specific files to your [global ignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) file instead.

composer.json
```diff
@@ -12,18 +12,18 @@
     }
   ],
   "require": {
-    "php": ">=5.5.9",
+    "php": "^7.1",
     "monolog/monolog": "^1.0",
-    "symfony/console": "^3.0",
```
phansys: Supporting version 4 of Symfony should not imply to drop support for version 3, since the features provided by version 3 are still compatible with the current codebase. Please, update the constraints to support both versions instead (`^3.0|^4.0`).

.travis.yml
```diff
@@ -1,8 +1,5 @@
 language: php
 php:
-  - '5.5'
```
phansys: Support for these versions should not be dropped.

unherz: Its right to drop support for php versions <= 7.0 because they aren't in active support anymore.  Even if 5.6 and 7.0 get security fixes, reach their end of life on Dec 2018.
See [php.net/supported-versions](http://php.net/supported-versions.php) for more.

bzoks: I think that since this is Symfony-related, version constraints should be aligned with constraints Symfony enforces. If this version supports Symfony 3.4, it should support same php versions as Symfony 3.4 (5.5+)

phansys: >Its right to drop support for php versions <= 7.0

IMO, it dependens which version this PR will affect. If `nostromo` is meant for `3.x` branch and `master` will alias `4.x`, I agree that we can accept non BC changes like this. If that is the case, I'll be happy to submit a PR for the current latest stable version (`nostromo`) allowing symfony 4 support.

andres-montanez:
Hi all, thanks for the PR, but this will brake a lot of things.
Symfony 4 will be supported in Magallanes 4.
Regards.

bzoks:
@andres-montanez , FYI, this fork works great with Symfony 4: https://github.com/sheeep/Magallanes/tree/features/symfony-4-compatibility

PR #414, opened 2018-01-26
Added sleep task to allow delaying execution of other tasks

Toflar:
Sometimes that's helpful, like if you have to wait a few moments until e.g. the `current` symlink is ready and the server was able to update internally etc.


andres-montanez (approved):

andres-montanez:
Thanks @Toflar , this is going to be added to v5!

PR #416, opened 2018-03-16
[FEATURE] allow "local deployments"

bbak:
If the target host is "localhost", then skip building a SSH command and just execute a command locally. This should allow "local deployments" - e.g. to deploy to a staging or testing environment on localhost.


andres-montanez:
Sounds good, but this will need more work too, what if I use 127.0.0.1?
What should we do with the defined user?
Doing the SSH is not a big overhead.

Issue #366, opened 2017-03-15
[Commands] Custom Commands


nexxome:
Is it possible to use custom commands or run single tasks only? For example to only update database schema on a specific environment

andres-montanez:
Hi, currently that's not possible and is not on the road map.
Following your example, if you want to update the db schema you should run the symfony command for that (or corresponding framework's console).

Regards.

PR #420, opened 2018-04-06
Event subscribers

Toflar:
There were already other issues about [registering custom commands](https://github.com/andres-montanez/Magallanes/issues/366) and I also [submitted a PR](https://github.com/andres-montanez/Magallanes/pull/402) for an ssh command which was declined and I have no possibility to add it on my own at the moment.

So I thought I'll come up with a simple concept that allows us to provide events for whatever extension point you may think of. The one that supports registering custom commands is just the first one but it serves as a useful example right away.

So subscribers are registered like so:

```yaml
magephp:
    subscribers:
        - Vendor\MageExtension\SuperEventSubscriber
```

All that class has to do is to implement the Symfony `EventSubscriberInterface` and register to any event we will add in the future.
The `Runtime` now has a `getEventDispatcher()` allowing to dispatch events everywhere within the application. You can do it in built-in tasks yourself or one can do it in custom tasks for whatever reason. This concept provides a lot of flexibility.

Using this concept, we can easily register a new command. So if I wanted to add my SSH command, my `.mage.yml` would look like so:

```yaml
magephp:
    subscribers:
        - Vendor\MageExtension\SshCommandProviderEventSubscriber
```

```php
class SshCommandProviderEventSubscriber implements EventSubscriberInterface
{
    public function registerCommands(RegisterCommandsEvent $event)
    {
        $event->addCommand(new SshCommand());
    }

    public static function getSubscribedEvents(): array
    {
        return [
            RegisterCommandsEvent::EVENT_NAME => ['registerCommands'],
        ];
    }
}
```

Then I can use `vendor/bin/mage ssh` 😄 

No tests yet, I would like to know your thoughts about the concept!

### DIFF
--- a/src/Event/RegisterCommandsEvent.php
+++ b/src/Event/RegisterCommandsEvent.php
@@ -20,7 +20,7 @@ class RegisterCommandsEvent extends Event
     /**
      * @var Command[]
      */
-    private $commands;
+    private $commands = [];
 
     /**
      * @return Command[]
--- a/src/Event/RegisterCommandsEvent.php
+++ b/src/Event/RegisterCommandsEvent.php
@@ -1,0 +1,56 @@
+<?php
+/*
+ * This file is part of the Magallanes package.
+ *
+ * (c) Andrés Montañez <andres@andresmontanez.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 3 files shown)


andres-montanez:
This is a bit overkill for the feature set of Mage.

PR #425, opened 2018-06-12
Add optional symlink

agalic89:


andres-montanez:
Thanks @agalic89 , this new option will be available on v5!

Issue #306, opened 2016-09-05
Windows problem


matiux:
I have a problem to deploy from windows.

I get this error when tar runs:

tar<child>: Cannot connect to C: resolve failed  

I have read that the reason is that tar interprets colons (:) in file names as meaning it is a file on another machine. tar has the --force-local option that solve this problem if I run the commnad manually in terminal

But how can I apply this option to magellan task?

andres-montanez:
Hi, Version 1 of Magallanes has long reached it's end of life cycle, so no further changes will be integrated to it, focus is for the next next mayor version.
Many thanks!

fritzmg:
It can be fixed by using
```
tar_create: cfzop --force-local
```
for example (_cfzop_ is the default for `tar_create`).

PR #427, opened 2018-09-10
Use fallback for posix_getpwuid on Windows

fritzmg:
With one other adjustment (see https://github.com/andres-montanez/Magallanes/issues/306#issuecomment-419921293) Magallanes seems to work without an issue under Windows, at least under the Windows Git Bash.

The only change I had to make in the source of Magallanes is to provide this fallback for `posix_getpwuid`, which is of course not present on the Windows platform.

### DIFF
--- a/src/Runtime/Runtime.php
+++ b/src/Runtime/Runtime.php
@@ -517,6 +517,11 @@ class Runtime
      */
     public function getCurrentUser()
     {
+        // Windows fallback
+        if (!function_exists('posix_getpwuid')) {
+            return getenv('USERNAME') ?: '';
+        }
+
         $userData = posix_getpwuid(posix_geteuid());
         return $userData['name'];
     }


tfritsch-km:
any chance this PR would be merged ?

fritzmg:
@tfritsch-km the project does not seem to be maintained anymore. Also I should still change the tests, so that this change is covered.

fritzmg:
@tfritsch-km my fork now contains a branch that includes all windows fixes so far: https://github.com/fritzmg/Magallanes/tree/windows-fixes

You could add that as a repository in your global `composer.json` and then require `"andres-montanez/magallanes": "dev-windows-fixes"`.

iannacone:
the alternatives are:
- `exec('whoami')` which i totally do not recommend for the high permissions requirement
- make the `getEnvOption` 2nd argument lazy loaded (that value is not always really needed), but this change requires some work

btw i honestly do not understand why the coverage decrease of the 0.1%.
imo under windows it throws an exception which means as now it is not cover 100% all the cases.
could you please be so gentle and explicit share what cases do not meet the requirements?

andres-montanez:
Hi @iannacone , the coverage decreases because there is no test entering inside the if added for the feature.
Magallanes has never been too kind with windows, however, I will take a look at this in the next weeks to improve compatibility.

Take care!

andres-montanez (approved):

andres-montanez:
Thanks @fritzmg , this improvement will be available on v5!

PR #429, opened 2018-11-20
[RFC] do not use PHP_EOL for remote data

### LINKED ISSUE #428: v4: does not delete old releases?
I have configured an environment with `releases: 2`. However, after successfully deploying a new release, old releases are never actually deleted. Currently there are already 7 releases in the `releases` directory - none of the old ones were ever deleted. The console output says the clean up is run successfully:
```
    Starting Post Release tasks on host […]:
        Running [Release] Cleaning up old Releases ... OK
    Finished 1/1 tasks for Post Release.
```
This task immediately finishes - which is odd anyway, because there would be a lot of files to delete and that would take at least a little bit of time.

fritzmg:
This fixes #428 

Currently Magallanes uses `PHP_EOL` for every instance where it wants to either parse or write a new 'new line character'.

However, when Magallanes is used on Windows, this breaks things like the `CleanupTask` (see #428) - if the remote host is a Unix machine (which it usually is).

The proposed solution isn't perfect and simply assumes, that the remote host will always be a Unix machine.

May be the __remote__ EOL character should be made configurable in the `.mage.yml` and simply defaults to `"\n"` otherwise. What do you think?

### DIFF
--- a/src/Command/BuiltIn/Releases/ListCommand.php
+++ b/src/Command/BuiltIn/Releases/ListCommand.php
@@ -97,7 +97,7 @@ class ListCommand extends AbstractCommand
 
                     $releases = [];
                     if (trim($process->getOutput()) != '') {
-                        $releases = explode(PHP_EOL, trim($process->getOutput()));
+                        $releases = explode("\n", trim($process->getOutput()));
                         rsort($releases);
                     }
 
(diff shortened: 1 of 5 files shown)


andres-montanez (approved):

andres-montanez:
Thanks @fritzmg , it makes sense.
I would have prefer to define a constant inside Magallanes and define it with the `"\n"`, but let's roll with it as is.

Thanks for your time and effort!

PR #434, opened 2019-05-01
fix tar_create command on Windows

fritzmg:
Under Windows, we need the `--force-local` argument for `tar_create` by default. Otherwise, due to the `:` in absolute Windows file paths, tar thinks that the path is on another machine - and you have to override this using `--force-local` (see also https://github.com/Microsoft/azure-pipelines-tool-lib/issues/34 for example).

Imho it is important to integrate this into Magallanes, so that a `.mage.yml` can be committed to a git repository and then used by Linux and Windows users alike. Otherwise if you put
```yml
tar_create: --force-local -c -z -p -f
```
into your `.mage.yml` it will not work for *nix users anymore - and without it it won't work for Windows users.


andres-montanez:
Hi @fritzmg , thank you very much for your time and effort, and thinking about the compatibility issues.
I will merge this PR and release it soon.
Thank you again!

PR #436, opened 2019-08-25
Make CopyTask scp command timeout configurable

artur-tud:
**Use case:**
Deploying using releases will copy the tar archive to the remote server with "scp command" with a timeout maximum of 300 seconds .

**Problem:**
If you have limited upload bandwidth and/or big tar archives, the command will run into a timeout after 300 seconds.
$process = $this->runtime->runLocalCommand($cmdCopy, 300);

**Solution:**
Make the timeout for the CopyTask (scp upload command) configurable in the .mage.yml file. The default value can still be 300 seconds.

### DIFF
--- a/src/Runtime/Runtime.php
+++ b/src/Runtime/Runtime.php
@@ -471,6 +471,10 @@ class Runtime
             $sshConfig['flags'] = '-q -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no';
         }
 
+        if (!array_key_exists('timeout', $sshConfig)) {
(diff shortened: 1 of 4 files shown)


andres-montanez:
Thanks man, I will review it when I'm able, probably port it to v3 too.
Regards.

nntoan:
Hi @andres-montanez , any updates on this? It would be very useful especially in case having trouble with ISP which slow down your internet speed.

andres-montanez (approved):

andres-montanez:
Thanks everyone for their time and effort!
I will be releasing this soon.

PR #438, opened 2019-10-30
Update README.md

tomasliubinas:
By default you probably would never want to run this tool on production? In this case it would be better to require it for --dev dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 Simply add the following dependency to your project’s composer.json file:
 
 ```json
-    "require": {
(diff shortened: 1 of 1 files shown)


phansys:
A great improvement should aim the user to use the command instead of edit the file manually:
```
composer require --dev andres-montanez/magallanes
```
Which should resolve the most recent compatible release.

andres-montanez:
Thanks everyone!
@tomasliubinas makes total sense, I will update it on the doc pages too!
Thanks everyone!!

PR #454, opened 2022-03-12
[FEATURE] Symfony 6 compatibility

MaxBoeh:
Added support for Symfony 6, so Magallanes can be used in Symfony 6 projects.

Some why's:
- Dropped support for php <8.0 because of symfony/process 6 declaring "static" as return type on a method that is overwritten bei Magallanes
- Dropped Symfony 4 from list of compatible Symfony versions, as at least Mage\Tests\Runtime\ProcessMockup::setTimeout is no longer compatible with Symfony\Component\Process\Process::setTimeout. 
- Although the contribution guide says one should start branches from `nostromo`, I've started from `master` as the former is quite dated.

Running the tests with the latest Symfony 5 and Symfony 6 packages works fine.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,12 +14,12 @@
   "require": {
     "php": "^8.0",
     "monolog/monolog": "~1.11 | ^2.0",
-    "symfony/console": "^4.0 | ^5.0 | ^6.0",
-    "symfony/filesystem": "^4.0 | ^5.0 | ^6.0",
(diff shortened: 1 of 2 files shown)


andres-montanez:
Hi, thanks for this!!
However, this may brake some installations with older PHP versions.
I'll take this into the next version for Symfony 6.

andres-montanez:
The upcoming version of Magallanes (V5), will be ported to Symfony 6 and with strict typing for PHP 8.

PR #458, opened 2022-04-15
Simplify workflow by leveraging build matrix

iambrosi:
This allows to reuse the workflow syntax for a job, tweaking small parts via configuration.


andres-montanez:
Thanks @iambrosi , this is awesome!

PR #459, opened 2022-04-18
Fix PHP version used by testing workflow

iambrosi:
The test workflow use the action `php-actions/composer` to install the project dependencies, specifying the PHP version that composer should use, but this action does not configure the test runner to continue using the desired PHP version.  This issue can be seen in [this example run](https://github.com/iambrosi/Magallanes/runs/6057929145?check_suite_focus=true), which simply added `php --version` before running PHPUnit:

image

The fix is to use the action `shivammathur/setup-php@v2` to setup the proper PHP version in the test runner, replacing `php-actions/composer` which only ensured using the desired PHP version while installing the project dependencies.

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -19,13 +19,13 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Composer install
-        uses: php-actions/composer@v5
+      - name: Setup PHP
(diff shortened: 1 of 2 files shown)


iambrosi:
Ping @andres-montanez for visibility

andres-montanez:
Hi @iambrosi , thank you a lot! ❤️

